PR #16, opened 2025-03-14
Cluster operations

rahulbats:
i added list clusters tool, this is required by a client i am working with


src/print-md-schema.ts
eddyv, line 1: revert, we have this class in here to help generate the markdown on the `README.md` :)

src/confluent/tools/tool-factory.ts
```diff
@@ -20,6 +22,7 @@ import { SearchTopicsByTagHandler } from "@src/confluent/tools/handlers/search/s
 import { ToolName } from "@src/confluent/tools/tool-name.js";
 import { DeleteConnectorHandler } from "./handlers/connect/delete-connector-handler.js";
 import { SearchTopicsByNameHandler } from "./handlers/search/search-topics-by-name-handler.js";
+import { ListClustersHandler } from "./handlers/clusters/list-clusters-handler.js";
```
eddyv, line 25: nit - fix paths
```suggestion
import { DeleteConnectorHandler } from "@src/confluent/tools/handlers/connect/delete-connector-handler.js";
import { SearchTopicsByNameHandler } from "@src/confluent/tools/handlers/search/search-topics-by-name-handler.js";
import { ListClustersHandler } from "@src/confluent/tools/handlers/clusters/list-clusters-handler.js";
```

eddyv (changes requested):
Thank you for your contribution!

What I may suggest before merging is to refactor the implementations of each handler to look a little bit more like https://github.com/confluentinc/mcp-confluent/blob/main/src/confluent/tools/handlers/connect/list-connectors-handler.ts

It should help consolidate there are some convenience methods and fetch clients so you don't need to inject auth headers yourself or the response text parsing.

Let me know if you need any help.

rahulbats:
accepted your change

eddyv (approved):

PR #22, opened 2025-04-25
Clean up README

tlberglund:
The README referenced a thing called "Confluent Kafka." Apache Software Foundation trademark guidelines do not allow us to use this phrase. :) 

My proposed modification might be too simple, but I'm not sure what the difference is between our version of Kafka (i.e., what we run in Confluent Cloud), and Confluent Cloud itself, which remains in the suggested version of the README.


eddyv (approved):
Ty!

PR #34, opened 2025-05-13
Get configurations for a given topic

mcolomerc:
Adds the tool to allow get topic configurations.

 
Example: 

Prompt: Get topic_11 configuration

Output: 

Here is the configuration for topic_11 in the source_cluster:

Partitions: 12
Replication factor: 3
cleanup.policy: delete
compression.type: producer
retention.ms: 604800000 (7 days)
max.message.bytes: 2097164
min.insync.replicas: 2
segment.bytes: 104857600

Let me know if you need more details or want to see the full configuration!


eddyv (approved):
lgtm, thank you!

PR #91, opened 2026-03-13
update CI pipeline to include static analysis and testing

shouples:
Basic expansion on the existing pipeline to fail early for any TS compilation issues or linting (which should be caught by the existing precommit hook), followed by a placeholder Test block to be integrated in a follow-up PR.
image

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -20,16 +20,17 @@ global_job_config:
     commands:
       - sem-version node 22
       - checkout
-      - npm ci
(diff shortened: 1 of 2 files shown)


service.yml
```diff
@@ -12,6 +12,7 @@ renovatebot:
 semaphore:
   enable: true
   pipeline_enable: false
+  status_level: block
```
shouples: Doesn't do anything in this branch, but once merged will give us block-level status as individual GitHub checks

flippingbits (approved):
🎉

PR #142, opened 2026-04-07
docs: update external docs links

shouples:
Small follow-up to #130 since some external links were returning 404 pages


jlrobins (approved):
Verified all changed links lead to MCPy pages.

PR #146, opened 2026-04-09
Have each tool register its read-only, create-update, or destructive delete intent

### LINKED ISSUE #145: Annotate the read-only-ness of every tool.
All 50 tools are registered without MCP spec annotations (destructiveHint, readOnlyHint). AI clients like Claude Desktop and Cursor have no signal to distinguish delete-topics from list-topics. A prompt injection can trigger destructive operations (delete-topics, delete-schema with permanent:true, DROP TABLE via create-flink-statement) with the same friction as read-only operations - no confirmation prompt from the AI client.

Add MCP spec tool annotations to each tool registration - readOnlyHint:true for read tools, destructiveHint:true for delete/drop tools.

New unit test proving that _every_ tool annotation includes a readOnly or destructive hint.

Github issues shadow of MATRIX-1505.

jlrobins:
Adds [MCP tool annotations](https://modelcontextprotocol.io/specification/2025-11-25/schema#toolannotations) (`readOnlyHint`, `destructiveHint`) to all 52 tools, enabling AI clients (Claude Desktop, Cursor) to distinguish between safe read operations (`list-topics`, `consume-messages`) and destructive operations (`delete-topics`, `delete-schema --permanent`, `delete-connector`).

Without these annotations, prompt injection could trigger destructive actions with the same friction as read-only queries.

Implements three annotation constants (`READ_ONLY`, `CREATE_UPDATE`, `DESTRUCTIVE`) referenced by all tool handlers, with unit tests validating both correct constant usage and adherence to naming conventions (tools prefixed with `delete-`/`remove-` use `DESTRUCTIVE`, `create-`/`update-`/`add-` use `CREATE_UPDATE`, etc.).

Closes #145.

### DIFF
--- a/src/confluent/tools/base-tools.ts
+++ b/src/confluent/tools/base-tools.ts
@@ -2,7 +2,24 @@ import { ClientManager } from "@src/confluent/client-manager.js";
 import { CallToolResult } from "@src/confluent/schema.js";
 import { ToolName } from "@src/confluent/tools/tool-name.js";
 import { EnvVar } from "@src/env-schema.js";
+import { ToolAnnotations } from "@modelcontextprotocol/sdk/types.js";
 import { ZodRawShape } from "zod";
+
+/**
+ * Standard MCP tool annotations.
+ * Tools should reference these constants rather than creating ad-hoc instances.
+ */
+export const READ_ONLY: ToolAnnotations = {
+  readOnlyHint: true,
+} as const;
+
+export const CREATE_UPDATE: ToolAnnotations = {
+  destructiveHint: false,
+} as const;
+
+export const DESTRUCTIVE: ToolAnnotations = {
+  destructiveHint: true,
+} as const;
 
 export interface ToolHandler {
   handle(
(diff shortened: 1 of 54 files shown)


src/confluent/tools/base-tools.ts
```diff
@@ -2,8 +2,25 @@ import { ClientManager } from "@src/confluent/client-manager.js";
...
+  readOnlyHint: true,
+} as const;
+
+export const CREATE_UPDATE: ToolAnnotations = {
+  destructiveHint: false,
+} as const;
+
+export const DESTRUCTIVE: ToolAnnotations = {
+  destructiveHint: true,
+} as const;
```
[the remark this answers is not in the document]
jlrobins: Revised.

src/confluent/tools/base-tools.ts
```diff
@@ -2,8 +2,25 @@ import { ClientManager } from "@src/confluent/client-manager.js";
...
```
shouples, line 24: Based on https://modelcontextprotocol.io/specification/2025-11-25/schema#toolannotations, should these also include `readOnlyHint: false`?
> (This property is meaningful only when readOnlyHint == false)

jlrobins: Done. I think that that '==' and not '===' is truthy, and that if left undefined they'd == as false, but added for exactness.

shouples (approved):

Issue #182, opened 2026-04-15
Add auto-refresh for managing token lifecycle


bcscc:
Add a background refresh loop for token store (#181). Should also include refresh callback that manages token lifecycle (expiry removal, near-expiry refresh).

PR #185, opened 2026-04-15
feat(oauth): implement token exchange chain

### LINKED ISSUE #180: Implement the token exchange chain
Implement auth code, CP, and DP token exchanges plus refresh token grant expiry tracking

bcscc:
### Summary
- Add stubbable `nodeFetch` and `nodeCrypto` wrappers in `node-deps.ts` for test isolation
- Add PKCE code verifier/challenge generation and opaque token utilities (`crypto-utils.ts`)
- Implement full Confluent token chain: auth code → ID token → control plane → data plane (`token-chain.ts`)
- Implement refresh token grant for background token renewal
- bound each token-chain fetch with a 30s timeout 
- 14 unit tests covering all exchange functions, error handling, and chain orchestration

Closes #180

### PR Stack
- #183
  - (this) #185
    - #187 
      - #193

### DIFF
--- a/src/confluent/node-deps.ts
+++ b/src/confluent/node-deps.ts
@@ -2,6 +2,7 @@
 import { Analytics } from "@segment/analytics-node";
 import { TELEMETRY_WRITE_KEY } from "@src/build-config.js";
 import envProxy from "@src/env.js";
+import { randomBytes } from "node:crypto";
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { arch, homedir, platform, release } from "node:os";
 import { join } from "node:path";
@@ -12,3 +13,5 @@ export const os = { homedir, platform, release, arch };
 export const path = { join };
 export const segment = { Analytics };
 export const config = { env: envProxy };
+export const nodeFetch = { fetch: globalThis.fetch };
+export const nodeCrypto = { randomBytes };
--- a/src/confluent/oauth/crypto-utils.test.ts
+++ b/src/confluent/oauth/crypto-utils.test.ts
@@ -11,7 +11,7 @@ describe("oauth/crypto-utils.ts", () => {
       const token = generateOpaqueToken();
 
       expect(token).toHaveLength(64);
-      expect(token).toMatch(/^[0-9a-f]{64}$/);
+      expect(Buffer.from(token, "hex")).toHaveLength(32);
     });
 
     it("should generate unique tokens on each call", () => {
@@ -26,9 +26,7 @@ describe("oauth/crypto-utils.ts", () => {
     it("should return a base64url-encoded string of appropriate length", () => {
       const verifier = generateCodeVerifier();
 
-      // 32 bytes → 43 base64url characters (no padding)
-      expect(verifier.length).toBeGreaterThanOrEqual(43);
-      expect(verifier).toMatch(/^[A-Za-z0-9_-]+$/);
+      expect(Buffer.from(verifier, "base64url")).toHaveLength(32);
     });
   });
 
@@ -37,9 +35,7 @@ describe("oauth/crypto-utils.ts", () => {
       const verifier = generateCodeVerifier();
       const challenge = generateCodeChallenge(verifier);
 
-      // SHA-256 → 32 bytes → 43 base64url characters
-      expect(challenge.length).toBe(43);
-      expect(challenge).toMatch(/^[A-Za-z0-9_-]+$/);
+      expect(Buffer.from(challenge, "base64url")).toHaveLength(32);
     });
 
     it("should produce different challenges for different verifiers", () => {
--- a/src/confluent/oauth/crypto-utils.test.ts
+++ b/src/confluent/oauth/crypto-utils.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it } from "vitest";
+import {
+  generateOpaqueToken,
+  generateCodeVerifier,
+  generateCodeChallenge,
+} from "@src/confluent/oauth/crypto-utils.js";
+
+describe("oauth/crypto-utils.ts", () => {
+  describe("generateOpaqueToken", () => {
+    it("should return a 64-character hex string", () => {
+      const token = generateOpaqueToken();
+
+      expect(token).toHaveLength(64);
+      expect(token).toMatch(/^[0-9a-f]{64}$/);
+    });
+
+    it("should generate unique tokens on each call", () => {
+      const token1 = generateOpaqueToken();
+      const token2 = generateOpaqueToken();
+
+      expect(token1).not.toBe(token2);
+    });
+  });
+
+  describe("generateCodeVerifier", () => {
+    it("should return a base64url-encoded string of appropriate length", () => {
+      const verifier = generateCodeVerifier();
+
+      // 32 bytes → 43 base64url characters (no padding)
+      expect(verifier.length).toBeGreaterThanOrEqual(43);
+      expect(verifier).toMatch(/^[A-Za-z0-9_-]+$/);
+    });
+  });
+
+  describe("generateCodeChallenge", () => {
+    it("should return a base64url-encoded SHA-256 hash of the verifier", () => {
+      const verifier = generateCodeVerifier();
+      const challenge = generateCodeChallenge(verifier);
+
+      // SHA-256 → 32 bytes → 43 base64url characters
+      expect(challenge.length).toBe(43);
+      expect(challenge).toMatch(/^[A-Za-z0-9_-]+$/);
+    });
+
+    it("should produce different challenges for different verifiers", () => {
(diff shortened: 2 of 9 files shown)


src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const now = Date.now();
+
+  return {
+    refreshToken: auth0Response.refresh_token,
+    refreshTokenExpiresAt: now + REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS,
+    controlPlaneToken: cpResponse.token,
+    controlPlaneExpiresAt: now + CONTROL_PLANE_TOKEN_LIFETIME_MS,
+    dataPlaneToken: dpResponse.token,
+    dataPlaneExpiresAt: now + DATA_PLANE_TOKEN_LIFETIME_MS,
```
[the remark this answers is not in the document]
bcscc: Since these TTL values are known, simpler to compute client side than adding parsing logic and achieves the same result.

src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,208 @@
...
+const REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS = 8 * 60 * 60 * 1000;
+
+export interface TokenChainResult {
+  refreshToken: string;
+  refreshTokenExpiresAt: number;
+  controlPlaneToken: string;
+  controlPlaneExpiresAt: number;
+  dataPlaneToken: string;
+  dataPlaneExpiresAt: number;
+}
```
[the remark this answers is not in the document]
bcscc: Both serve different layers, TokenChainResult represents what the exchange chain returns, ConfluentTokenSet is what gets stored so will keep seperate.

src/confluent/oauth/crypto-utils.ts
```diff
@@ -0,0 +1,25 @@
+import { createHash, randomBytes } from "node:crypto";
+
```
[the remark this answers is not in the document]
bcscc: need imports to verify output format

src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,208 @@
...
+  const dpResponse = await exchangeControlPlaneForDataPlaneToken(
+    auth0Config.apiUrl,
+    cpResponse.token,
+  );
+
+  const now = Date.now();
+
+  return {
+    refreshToken: auth0Response.refresh_token,
+    refreshTokenExpiresAt: now + REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS,
```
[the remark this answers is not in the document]
bcscc: Updated to track both absolute and idle TTL. Absolute will remain the same based on the initial login, while idle will be updated when refresh token is used

src/confluent/oauth/crypto-utils.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      expect(token1).not.toBe(token2);
+    });
+  });
+
+  describe("generateCodeVerifier", () => {
+    it("should return a base64url-encoded string of appropriate length", () => {
+      const verifier = generateCodeVerifier();
+
+      // 32 bytes → 43 base64url characters (no padding)
```
jlrobins: If is supposed to be b64-encoded, then perhaps better to prove that it can be b64-decoded (into expected length at least) than just regex + supposed-b64-encoded length tested?

And likewise in all of the subsequent tests over what should be b64-encoded strings. Prove decodable and then make assertions over the decoded results.

src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,224 @@
+import type {
+  Auth0Config,
+  Auth0TokenResponse,
+  ControlPlaneTokenResponse,
+  DataPlaneTokenResponse,
+} from "@src/confluent/oauth/types.js";
+import { nodeFetch } from "@src/confluent/node-deps.js";
+import { logger } from "@src/logger.js";
+
+/** 5 minutes in milliseconds — control plane token lifetime */
```
jlrobins, line 10: Question: Having these values be hardcoded within the codebase feels a little off to me --- can't the expiration times be obtained / derived _from the tokens themselves_ so that we'd not ever be misaligned?

That is, to have the expiration times derived from _decoding_ the JWTs?

https://www.npmjs.com/package/jwt-decode

src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,224 @@
...
+/** 10 minutes in milliseconds — data plane token lifetime */
+const DATA_PLANE_TOKEN_LIFETIME_MS = 10 * 60 * 1000;
+
+/** 8 hours in milliseconds — refresh token absolute lifetime from original login */
+const REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS = 8 * 60 * 60 * 1000;
+
+/** 4 hours in milliseconds — refresh token idle timeout, resets on each rotation */
+const REFRESH_TOKEN_IDLE_LIFETIME_MS = 4 * 60 * 60 * 1000;
+
+export interface TokenChainResult {
```
jlrobins, line 31: Docstring on the use / intent of `TokenChainResult` overall please.

src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  return (await response.json()) as DataPlaneTokenResponse;
+}
+
+/**
+ * Uses a refresh token to obtain new Auth0 tokens, then derives CP and DP tokens.
+ * This is used by the background refresh loop.
+ */
+export async function refreshTokenChain(
```
jlrobins, line 150: Are there any considerations on how to ensure that this is happening at most once at any time in the MCP process (and then with a timeout?)

Something like an async lock guard, or a module-level reference to a pending promise + timeout pair, then this function deciding to return the existing promise or to make a new one?

src/confluent/oauth/token-chain.ts
```diff
@@ -0,0 +1,224 @@
...
+    refresh_token: refreshToken,
+  });
+
+  const response = await nodeFetch.fetch(tokenUrl, {
+    method: "POST",
+    headers: { "Content-Type": "application/x-www-form-urlencoded" },
+    body: body.toString(),
+  });
+
+  if (!response.ok) {
```
jlrobins: This repeated `if (!response.ok)` ... `await response.json() as T` is repeated and could be compressed into an interior utility generic function  over T and the core part of an exception message. Repeated.

_Cells interlinked within cells interlinked_
_Within cells interlinked_
_Within cells interlinked_
_Within cells interlinked_

src/confluent/node-deps.ts
```diff
@@ -12,3 +13,5 @@ export const os = { homedir, platform, release, arch };
 export const path = { join };
 export const segment = { Analytics };
 export const config = { env: envProxy };
+export const nodeFetch = { fetch: globalThis.fetch };
```
jlrobins, line 16: Is there any other `fetch` that would be in play in the codebase other than Node's? Why not just call this  shadowing symbol `fetch` like we're doing all the other things here?

jlrobins (changes requested):
Core functionality looks good. Main asks are for a little more specific tests, and then question about ability to obtain the expiration times _from the tokens themselves_ instead of the codebase making assumptions.

Would be perfectly happy to see the tests tightening done here in this PR, but can push the more invasive requests to a branch at the _tail end_ of this stack so that you won't have to do more rebasing and respelling than necessary as work through the existing stack of work.

src/confluent/oauth/token-chain.ts
bcscc: I was thinking since these lifetime values are known and unlikely to change unless the Auth0 client configs are manually changed, having these hard-coded keeps it a bit simpler and values clear. This is also the pattern in [ide-sidecar](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/resources/application.yml#L127) however happy to create an issue to do a tail-end follow-up if decoding the JWTs is preferred.

jlrobins (approved):
Thanks for the improvements.

src/confluent/oauth/token-chain.ts
bcscc: This function will only have one caller in the refresh loop when I make a PR for #182, so I'll include a reentrancy boolean to handle the "at most once" side. Added the timeout (30s) half here on each fetch so a hung Auth0/Confluent call can't stall the loop.

src/confluent/node-deps.ts
bcscc: then it'll look like fetch.fetch which I thought would be a bit confusing, however happy to rename if we want consistency

src/confluent/oauth/token-chain.ts
bcscc: Follow issue here: https://github.com/confluentinc/mcp-confluent/issues/192?reload=1

PR #193, opened 2026-04-17
feat(oauth): auto-refresh token lifecycle

### LINKED ISSUE #182: Add auto-refresh for managing token lifecycle

bcscc:
### Summary

Background refresh for the OAuth token store:

- Scans stored token sets; removes any whose refresh token has expired (8hr absolute or 4hr idle); skips any whose control-plane token is still fresh; otherwise refreshes.
- Each refresh runs in two atomic phases. Phase 1 rotates the Auth0 refresh token and persists it immediately (partial update). Phase 2 derives new control-plane and data plane tokens and persists them together. A phase-2 failure leaves the rotated refresh safely stored so the next tick can retry without bricking the session.
- Retry on phase-2 failure is deferred to #195, where it ships with transient vs non-transient error classification.
- Shared lifetime constants (CP/DP/refresh/window) with `DEFAULT_REFRESH_INTERVAL_MS` derived as `CP_LIFETIME - REFRESH_WINDOW`.
- `startAutoRefresh` convenience wires the callback into the store's refresh loop.
- Tightens `TokenChainResult` (absolute refresh expiry now required) and relaxes `TokenStore.update` to accept a partial field set.

Closes #182

### PR Stack
- #183
  - #185
    - #187
      - (this) #193

### DIFF
--- a/src/confluent/oauth/token-lifetimes.ts
+++ b/src/confluent/oauth/token-lifetimes.ts
@@ -1,0 +1,14 @@
+/** 5 minutes in milliseconds — control plane token lifetime */
+export const CONTROL_PLANE_TOKEN_LIFETIME_MS = 5 * 60 * 1000;
+
+/** 10 minutes in milliseconds — data plane token lifetime */
+export const DATA_PLANE_TOKEN_LIFETIME_MS = 10 * 60 * 1000;
+
+/** 8 hours in milliseconds — refresh token absolute lifetime from original login */
+export const REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS = 8 * 60 * 60 * 1000;
+
+/** 4 hours in milliseconds — refresh token idle timeout, resets on each rotation */
+export const REFRESH_TOKEN_IDLE_LIFETIME_MS = 4 * 60 * 60 * 1000;
+
+/** 4 minutes in milliseconds — default auto-refresh interval (1 min before CP expiry) */
+export const DEFAULT_REFRESH_INTERVAL_MS = 4 * 60 * 1000;
--- a/src/confluent/oauth/token-lifetimes.ts
+++ b/src/confluent/oauth/token-lifetimes.ts
@@ -10,5 +10,9 @@ export const REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS = 8 * 60 * 60 * 1000;
 /** 4 hours in milliseconds — refresh token idle timeout, resets on each rotation */
 export const REFRESH_TOKEN_IDLE_LIFETIME_MS = 4 * 60 * 60 * 1000;
 
-/** 4 minutes in milliseconds — default auto-refresh interval (1 min before CP expiry) */
-export const DEFAULT_REFRESH_INTERVAL_MS = 4 * 60 * 1000;
+/** 1 minute in milliseconds — refresh control plane token this much before expiry */
+export const CONTROL_PLANE_REFRESH_WINDOW_MS = 60 * 1000;
+
+/** Default auto-refresh interval — CP lifetime minus the refresh window. */
+export const DEFAULT_REFRESH_INTERVAL_MS =
+  CONTROL_PLANE_TOKEN_LIFETIME_MS - CONTROL_PLANE_REFRESH_WINDOW_MS;
--- a/src/confluent/oauth/token-refresh.ts
+++ b/src/confluent/oauth/token-refresh.ts
@@ -1,21 +1,106 @@
-import type { Auth0Config } from "@src/confluent/oauth/types.js";
+import type {
+  Auth0Config,
+  ConfluentTokenSet,
+} from "@src/confluent/oauth/types.js";
 import type { TokenStore } from "@src/confluent/oauth/token-store.js";
-import { refreshTokenChain } from "@src/confluent/oauth/token-chain.js";
+import {
+  exchangeControlPlaneForDataPlaneToken,
+  exchangeIdTokenForControlPlaneToken,
+  exchangeRefreshTokenForAuth0Tokens,
+} from "@src/confluent/oauth/token-chain.js";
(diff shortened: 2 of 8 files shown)


src/confluent/oauth/token-refresh.ts
```diff
@@ -0,0 +1,155 @@
...
+      const tokenSet = store.get(accessToken);
+      if (!tokenSet) continue;
+
+      const now = Date.now();
+
+      if (
+        now >= tokenSet.refreshTokenAbsoluteExpiresAt ||
+        now >= tokenSet.refreshTokenIdleExpiresAt
+      ) {
+        logger.info("Refresh token expired, removing token set");
```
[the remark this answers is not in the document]
bcscc: unnesscary

src/confluent/oauth/token-store.ts
```diff
@@ -39,22 +39,16 @@ export class TokenStore {
...
 
-    existing.refreshToken = fields.refreshToken;
-    existing.refreshTokenIdleExpiresAt = fields.refreshTokenIdleExpiresAt;
-    existing.controlPlaneToken = fields.controlPlaneToken;
-    existing.controlPlaneExpiresAt = fields.controlPlaneExpiresAt;
-    existing.dataPlaneToken = fields.dataPlaneToken;
-    existing.dataPlaneExpiresAt = fields.dataPlaneExpiresAt;
-
+    Object.assign(existing, fields);
     return true;
```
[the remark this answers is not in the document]
bcscc: unnecessary

src/confluent/oauth/token-refresh.ts
```diff
@@ -0,0 +1,155 @@
...
+export async function refreshTokenSet(
+  auth0Config: Auth0Config,
+  store: TokenStore,
+  accessToken: string,
+  tokenSet: ConfluentTokenSet,
+): Promise<void> {
+  // Phase 1: rotate the refresh token with Auth0. Auth0 invalidates the old
+  // token on success, so we persist the rotated token IMMEDIATELY before
+  // attempting CP/DP exchange — otherwise a CP/DP failure would leave the
+  // store with an already-burned refresh token and brick the session.
```
flippingbits, line 31: Nice 👍 

Could two or multiple token refresh attempts potentially overlap? If so, we should use locks to make sure that only a single refresh attempt is performed at a time.

src/confluent/oauth/token-refresh.ts
```diff
@@ -0,0 +1,155 @@
...
+  } catch (error) {
+    logger.error(
+      { error },
+      "Auth0 refresh failed; keeping existing stored token set for next tick",
+    );
+    return;
+  }
+
+  const rotationTime = Date.now();
+  const rotatedIdleExpiresAt = rotationTime + REFRESH_TOKEN_IDLE_LIFETIME_MS;
```
flippingbits, line 47: Do we need to cap this at the end of the absolute lifetime?

src/confluent/oauth/token-refresh.ts
```diff
@@ -0,0 +1,155 @@
...
+    );
+
+    const updated = store.update(accessToken, {
+      controlPlaneToken: cpResponse.token,
+      controlPlaneExpiresAt: rotationTime + CONTROL_PLANE_TOKEN_LIFETIME_MS,
+      dataPlaneToken: dpResponse.token,
+      dataPlaneExpiresAt: rotationTime + DATA_PLANE_TOKEN_LIFETIME_MS,
+    });
+
+    if (!updated) {
```
flippingbits, line 82: Fair nitpick from SonarQube.

src/confluent/oauth/token-lifetimes.ts
```diff
@@ -0,0 +1,18 @@
...
+export const DATA_PLANE_TOKEN_LIFETIME_MS = 10 * 60 * 1000;
+
+/** 8 hours in milliseconds — refresh token absolute lifetime from original login */
+export const REFRESH_TOKEN_ABSOLUTE_LIFETIME_MS = 8 * 60 * 60 * 1000;
+
+/** 4 hours in milliseconds — refresh token idle timeout, resets on each rotation */
+export const REFRESH_TOKEN_IDLE_LIFETIME_MS = 4 * 60 * 60 * 1000;
+
+/** 1 minute in milliseconds — refresh control plane token this much before expiry */
+export const CONTROL_PLANE_REFRESH_WINDOW_MS = 60 * 1000;
```
flippingbits, line 14: nit: 1 minute feels a bit high from my experience.

flippingbits (approved):
Great work! I have only a few clarifying/non-blocking questions.

src/confluent/oauth/token-refresh.ts
bcscc: Yes good catch will fix in a followup #237

src/confluent/oauth/token-refresh.ts
bcscc: Thanks will fix in the follow up #237

src/confluent/oauth/token-refresh.ts
bcscc: The background loop is already single-flight via the `tickInProgress` guard in `TokenStore.startRefreshLoop`, so two scheduler ticks can't overlap, so a single refresh shouldn't overlap with itself

src/confluent/oauth/token-lifetimes.ts
bcscc: updated in #237

src/confluent/oauth/token-refresh.ts
flippingbits: Thanks!

Issue #165, opened 2026-04-13
Define the `auth` YAML structure capable of modeling api key/secret


jlrobins:
Define the yaml config syntax for an api_key and api_secret bearing `auth` block, necessary for the migration of all of the rest of our existing tools.

  * New discriminated union substructure tree Zod validation. This type will be 'api_key'.  In the future, there will be other auth mechanisms to model, such as username + password, etc.
  * Becomes optional members of DirectConnectionConfig's`kafka` and `schema_registry` members.
  * Add in support for `consConfigFromEnv` for both Kafka and Schema Registry key + secrets from the corresponding env vars as found in Environment.
  * Tests, tests, tests. Let's start adding working example yaml config files in `test-fixtures/yaml_configs`, including preexisting forms plus also a couple of variants including api-key auth blocks. Write new test block driving loading of fixture files via un-stubbed `loadConfigFromYaml` and investigating the results.

Issue #162, opened 2026-04-13
Refactor the upper portion of main to allow to produce an MCPServerConfiguration from the minimum env vars


jlrobins:
Refactor things so that there is a parallel code path to construct an MCPServerConfiguration instance from environment variables only, starting with just the two naming Kafka bootstrap servers and the SR endpoint.

The end result should be parity between 'config from env vars' and 'config from yaml', at least for the existing subset of functionality configurable from yaml, clearing the way for for the end-goal of phase one to allow for _either_ env var config or yaml config parity. This will require greatly expanding the yaml dialect to cover at least the ~15+ env vars that contribute to the construction of `DefaultClientManager`.

Drive this codepath if the `--config` cmdline option is _not_ provided.

Initially discard the MCPServerConfiguration just as we already do, for now.

(First raise PR against main to clean review, then we'll cross-merge into a new long-lived integration branch for the rest of the phase one work which will include partially broken states).

PR #210, opened 2026-04-22
Cover the remainder of Kafka env vars and the telemetry env vars.

### LINKED ISSUE #205: Kafka cluster metadata + Telemetry
Theme: Kafka cluster identity (REST endpoint, cluster ID, env ID) alongside optional observability. These are distinct from "Kafka auth" (already done in #165) — they're metadata used by REST handlers and catalog resolver.

What changes:
  * models.ts: extend kafka block with rest_endpoint, cluster_id, env_id; add top-level telemetry block (endpoint?, auth?: AuthConfig)
  * env-config.ts: extend for KAFKA_REST_ENDPOINT, KAFKA_CLUSTER_ID, KAFKA_ENV_ID, TELEMETRY_ENDPOINT, TELEMETRY_API_KEY, TELEMETRY_API_SECRET (6 vars)
 * Tests + fixture YAML

jlrobins:
## Summary of Changes

- Extend `DirectConnectionConfig` and `directConnectionSchema` with a `telemetry` block containing `endpoint?: string` and `auth?: AuthConfig` (both optional, but the block is rejected if neither is present — same pattern as `confluent_cloud`). Telemetry auth is intentionally optional because `constructDefaultClientManager()` already falls back to `confluent_cloud` credentials at the point of use.
- Extend the `kafka` block with three new optional fields: `rest_endpoint: string`, `cluster_id: string`, and `env_id: string` (must start with `env-`). These fields are used by REST-based Kafka handlers and the catalog resolver — they are metadata about the cluster identity, distinct from the auth credentials added in #165. As a consequence, `bootstrap_servers` is now optional in both the TypeScript interface and the Zod schema; at least one of the four connectivity fields (`bootstrap_servers`, `rest_endpoint`, `cluster_id`, `env_id`) is required via a `.refine()`.
- A present-but-empty `telemetry: {}` block is rejected as degenerate, consistent with `confluent_cloud` and `tableflow`.
- Extend `consConfigFromEnv()` for six new env vars: `KAFKA_REST_ENDPOINT`, `KAFKA_CLUSTER_ID`, `KAFKA_ENV_ID`, `TELEMETRY_ENDPOINT`, `TELEMETRY_API_KEY`, `TELEMETRY_API_SECRET`. The "at least one block must be defined" guard now accepts `telemetry` as sufficient on its own.

### Going above and beyond: two refactors forced by accumulated scale

With six blocks now wired into `consConfigFromEnv()`, the function had grown into a chain of inline `if`-blocks — one per service, each with its own ad-hoc structure. Adding a seventh inline would have pushed ESLint's cognitive-complexity limit and made the function unreadable. Rather than just bolt on the new code, this PR refactors the function to a builder-loop architecture:

- Each service block is extracted into its own focused `buildXxxBlock(env)` function. Each builder returns `{ <blockKey>: { ...fields } } | null` — the key is owned by the builder, not the caller. The orchestrator reduces to a flat loop over the builder list and `Object.assign`.
- Zod's `.strict()` acts as the safety net against misspelled keys in builder output: `mcpConfigSchema.safeParse()` rejects any unknown key at every level of the document, so typos in the builder's returned key are caught at validation time rather than silently accepted.
- Two builders (`buildSchemaRegistryBlock`, `buildFlinkBlock`) return inline types with all fields optional rather than their corresponding named interface, because the interface has required fields that the builder sets conditionally — Zod then enforces those at runtime and surfaces the error in env-var terms via `humanizeEnvConfigPaths()`.
- Each builder carries a JSDoc with an equivalent YAML snippet using the `${VAR}` interpolation syntax, making it easy to see the env-to-YAML mapping without cross-referencing two files.

Separately, the inline sub-types embedded in `DirectConnectionConfig` were promoted to named exported interfaces (`KafkaDirectConfig`, `SchemaRegistryDirectConfig`, `ConfluentCloudDirectConfig`, `TableflowDirectConfig`, `TelemetryDirectConfig`, `FlinkDirectConfig`). With six blocks, the inline definitions had grown large enough to obscure the shape of `DirectConnectionConfig` itself, and the named types are now useful as return type annotations on the builder functions.

Finally, the test file for `consConfigFromEnv()` had accumulated four nearly-identical `baseXxxEnv` constants — each listing all 20+ env vars as `undefined` just to satisfy the `Pick` type, with minor differences. These were replaced with a single `envWith(overrides?)` helper that fills every key as `undefined` and spreads the provided overrides, cutting the test file roughly in half and making each test read as a statement of intent rather than a transcription of the full env var namespace.

The maximal YAML config is now:

```yaml
connections:
  production:
    type: direct
    confluent_cloud:
      endpoint: "https://api.confluent.cloud"
      auth:
        type: api_key
        key: "my-cloud-key"
        secret: "my-cloud-secret"
    kafka:
      bootstrap_servers: "broker.confluent.cloud:9092"
      rest_endpoint: "https://pkc-abc123.us-east-1.aws.confluent.cloud:443"
      cluster_id: "lkc-abc123"
      env_id: "env-xyz789"
      auth:
        type: api_key
        key: "my-kafka-key"
        secret: "my-kafka-secret"
    schema_registry:
      endpoint: "https://psrc-abc123.us-east-2.aws.confluent.cloud"
      auth:
        type: api_key
        key: "my-sr-key"
        secret: "my-sr-secret"
    tableflow:
      auth:
        type: api_key
        key: "my-tableflow-key"
        secret: "my-tableflow-secret"
    telemetry:
      endpoint: "https://api.telemetry.confluent.cloud"
      auth:
        type: api_key
        key: "my-telemetry-key"
        secret: "my-telemetry-secret"
    flink:
      endpoint: "https://flink.us-east-1.aws.confluent.cloud"
      auth:
        type: api_key
        key: "my-flink-key"
        secret: "my-flink-secret"
      environment_id: "env-abc123"
      organization_id: "org-xyz789"
      compute_pool_id: "lfcp-pool01"
      environment_name: "my-environment"
      database_name: "my-kafka-cluster"
```

The additional env vars now supported:

```
KAFKA_REST_ENDPOINT=https://pkc-abc123.us-east-1.aws.confluent.cloud:443
KAFKA_CLUSTER_ID=lkc-abc123
KAFKA_ENV_ID=env-xyz789
TELEMETRY_ENDPOINT=https://api.telemetry.confluent.cloud
TELEMETRY_API_KEY=my-telemetry-key
TELEMETRY_API_SECRET=my-telemetry-secret
```

Closes #205
Step of #162
Successor to #208

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes


src/config/env-config.ts
```diff
@@ -58,65 +70,256 @@ export function consConfigFromEnv(
...
+    !env.TELEMETRY_API_KEY &&
+    !env.TELEMETRY_API_SECRET
+  )
+    return null;
+  return {
+    telemetry: {
+      ...(env.TELEMETRY_ENDPOINT && { endpoint: env.TELEMETRY_ENDPOINT }),
+      ...apiKeyAuth(env.TELEMETRY_API_KEY, env.TELEMETRY_API_SECRET),
+    },
+  };
```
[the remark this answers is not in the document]
jlrobins: Good catch on the default. The correct fix per the pattern established in #207 (where CONFLUENT_CLOUD_REST_ENDPOINT was handled the same way) is to remove the default from env-schema.ts rather than building from raw process.env. The fallback to https://api.telemetry.confluent.cloud was already applied at the point of use in index.ts via ?? — the schema default was redundant and had the side effect you identified. Removed in the latest commit.

src/config/env-config.ts
```diff
@@ -58,65 +70,256 @@ export function consConfigFromEnv(
...
-    env.CONFLUENT_CLOUD_API_SECRET
-  ) {
-    connection.confluent_cloud = {
+    !env.CONFLUENT_CLOUD_REST_ENDPOINT &&
+    !env.CONFLUENT_CLOUD_API_KEY &&
+    !env.CONFLUENT_CLOUD_API_SECRET
+  )
+    return null;
+  return {
+    confluent_cloud: {
```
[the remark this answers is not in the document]
jlrobins: Same root cause as comment 1, same fix. The CONFLUENT_CLOUD_REST_ENDPOINT default was supposed to be removed from env-schema.ts in #207 (the PR description even says so), but the implementation didn't follow through. Removed now, with the ?? fallback moved to the point of use in index.ts alongside the existing TELEMETRY_ENDPOINT fallback.

src/env-schema.ts
```diff
@@ -222,8 +222,7 @@ const configSchema = z
       .string()
       .describe("Base URL for Confluent Cloud's REST API services")
       .trim()
-      .url()
-      .default("https://api.confluent.cloud"),
+      .url(),
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/config/env-config.test.ts
```diff
@@ -1,14 +1,50 @@
 import { consConfigFromEnv } from "@src/config/env-config.js";
 import { describe, expect, it } from "vitest";
 
+type EnvInput = Parameters<typeof consConfigFromEnv>[0];
+
+/** Returns a fully-populated env object with every known key set to undefined, overridden by the provided values. */
+function envWith(overrides: Partial<EnvInput> = {}): EnvInput {
+  return {
+    BOOTSTRAP_SERVERS: undefined,
```
shouples, line 9: Any reason these tests can't use `envFactory`? https://github.com/confluentinc/mcp-confluent/blob/b10071f706bfab694781ae8dd27416caa0089299/tests/factories/env.ts#L44-L51

src/config/env-config.ts
```diff
@@ -58,65 +70,256 @@ export function consConfigFromEnv(
...
+  // Each builder returns { <blockKey>: { ...fields } } or null — the key is owned
+  // by the builder, not the caller. See builder docstrings for the env vars each handles.
+  const builders = [
+    buildKafkaBlock,
+    buildSchemaRegistryBlock,
+    buildConfluentCloudBlock,
+    buildTableflowBlock,
+    buildTelemetryBlock,
+    buildFlinkBlock,
+  ];
```
shouples, line 82: Nice cleanup refactoring here! 👏

shouples (approved):
Just the one earlier comment about the `envFactory` duplication, otherwise LGTM

src/config/env-config.ts
jlrobins: Thanks! Yeah, adding two more here in the old way was urban sprawl. And this also forced fracking the big interface up, which was also a plus.

src/config/env-config.test.ts
jlrobins: envFactory sets CONFLUENT_CLOUD_REST_ENDPOINT and TELEMETRY_ENDPOINT to non-undefined defaults, which causes buildConfluentCloudBlock and buildTelemetryBlock to emit blocks unconditionally — silently injecting two connection blocks into every test and breaking assertions like expect(conn.confluent_cloud).toBeUndefined(). The local envWith() uses a clean all-undefined baseline so each builder trigger is fully under test control, which is a different semantic contract than envFactory's production-defaults baseline. 

So -- is not a pure drop in replacement. But you're def right we ought not have two of these things. #212  created.

Issue #226, opened 2026-04-22
Migrate tableflow handlers to enabledConnectionIds()


jlrobins:
Override `enabledConnectionIds()` in all `handlers/tableflow/` handlers with `connectionIdsWhere(connections, hasTableflow)`.

Because all 11 tableflow handlers share the same predicate, introduce an intermediate base class rather than duplicating the one-liner across every handler:

```typescript
// src/confluent/tools/handlers/tableflow/tableflow-tool-handler.ts
export abstract class TableflowToolHandler extends BaseToolHandler {
  enabledConnectionIds(runtime: ServerRuntime): string[] {
    return connectionIdsWhere(runtime.config.connections, hasTableflow);
  }
}
```

Then extend `TableflowToolHandler` instead of `BaseToolHandler` in every tableflow handler, and delete each handler's `getRequiredEnvVars()` implementation.

Update each handler's `enabledConnectionIds()` test to use `runtimeWith(envFactory(), { tableflow: { ... } })` (no env vars — predicate only), proving the handler no longer consults `runtime.env`.

Issue #225, opened 2026-04-22
Migrate search handlers to enabledConnectionIds()


jlrobins:
Override `enabledConnectionIds()` in all `handlers/search/` handlers with the appropriate `connectionIdsWhere(connections, hasX)` one-liner. Delete the corresponding `getRequiredEnvVars()` implementation from each handler.

Issue #224, opened 2026-04-22
Migrate schema handlers to enabledConnectionIds()


jlrobins:
Override `enabledConnectionIds()` in all `handlers/schema/` handlers with the appropriate `connectionIdsWhere(connections, hasX)` one-liner. Delete the corresponding `getRequiredEnvVars()` implementation from each handler.

Issue #223, opened 2026-04-22
Migrate metrics handlers to enabledConnectionIds()


jlrobins:
Override `enabledConnectionIds()` in all `handlers/metrics/` handlers with the appropriate `connectionIdsWhere(connections, hasX)` one-liner. Delete the corresponding `getRequiredEnvVars()` implementation from each handler.

PR #255, opened 2026-04-28
Add new predicate functions and ToolHandler enabledConnectionIds() lifecycle method

### LINKED ISSUE #227: Infrastructure: add connection-predicates.ts, enabledConnectionIds() shim, flip getToolHandlersToRegister()
The critical-path PR 0 for #173.

- Add `src/confluent/tools/connection-predicates.ts` with `ConnectionPredicate` type, named predicates (`hasKafka`, `hasKafkaRest`, `hasSchemaRegistry`, `hasConfluentCloud`, `hasFlink`, `hasTelemetry`, `hasTableflow`), and `connectionIdsWhere(connections, predicate)` helper.
- Add `enabledConnectionIds(connections: Record<string, ConnectionConfig>): string[]` to `BaseToolHandler` as a **non-abstract** method with a shim that delegates to `getRequiredEnvVars()`. All existing handlers inherit it automatically — no handler changes required in this PR.
- Flip `getToolHandlersToRegister()` to call `enabledConnectionIds()` instead of `getRequiredEnvVars()` directly.

Behaviour is unchanged after this PR. It unblocks the domain migration children (#216–#226).

jlrobins:
# Infrastructure: `connection-predicates.ts` + `enabledConnectionIds()` shim

## Summary of Summary

Create the new ToolHandler lifecycle method `enabledConnectionIds(runtime: ServerRuntime)` which will replace `getRequiredEnvVars()` and tools last common / primary use of knowledge of env vars. Do so by initially implementing `enabledConnectionIds()` in terms of the (only during this transition period) env var payload within `ServerRuntime` by then downcalling into in-every-handler-but-no-longer-treated-as-a public-method `getRequiredEnvVars()` (now is called only by the BaseTool's `enabledConnectionIds()`. Create the module for the predicate functions which will actually assist the real implementations of `enabledConnectionIds()` at either the tool-family level or individual tool handler implementations. See description of #227 and then #173 for more context.

## Summary

- **`src/confluent/tools/connection-predicates.ts`** (new) — `ConnectionPredicate` type and seven named predicate functions (`hasKafka`, `hasKafkaRest`, `hasSchemaRegistry`, `hasConfluentCloud`, `hasFlink`, `hasTelemetry`, `hasTableflow`) over `ConnectionConfig`, plus `connectionIdsWhere(connections, predicate)` helper that filters a connection map to IDs satisfying a predicate.
- **`src/confluent/tools/connection-predicates.test.ts`** (new) — full coverage of every predicate (true and false path each) and `connectionIdsWhere` (empty match, single match, multiple match).
- **`src/confluent/tools/base-tools.ts`** — `getRequiredEnvVars()` removed from the `ToolHandler` interface; it survives as an `abstract` method on `BaseToolHandler` only, serving as the internal hook for the shim. `enabledConnectionIds(runtime: ServerRuntime): string[]` added to both the interface (the new public contract for tool enablement) and `BaseToolHandler` (concrete shim implementation that delegates to `getRequiredEnvVars()` against `runtime.env`). `EnvVar` disappears from the public tool surface entirely. Soon-to-be-removed-entirely `getRequiredEnvVars()` now marked `protected` due to only caller now being in-class-family `enabledConnectionIds()`.
- **`src/confluent/tools/base-tools.test.ts`** (new) — direct unit tests for `enabledConnectionIds()` via a minimal `StubHandler` subclass; covers all-vars-present, one-missing, all-missing, and no-requirements cases.
- **`src/index.ts`** — `getToolHandlersToRegister()` body flipped to call `handler.enabledConnectionIds(runtime)` instead of `handler.getRequiredEnvVars()` against `runtime.env`; `EnvVar` import removed. Disabled-tool log message updated to reflect the connection-centric framing: `"Tool X disabled; no connections satisfy its requirements"`.
- **`src/index.test.ts`** — `fakeHandler` updated to implement `enabledConnectionIds()` (removing the now-absent `getRequiredEnvVars`); local `runtimeWith()` helper removed in favour of the shared factory below.
- **`src/confluent/tools/handlers/kafka/list-topics-handler.test.ts`** — `enabledConnectionIds()` describe block added as a concrete real-handler proof that the shim wires correctly through a genuine `getRequiredEnvVars()` declaration.
- **`tests/factories/runtime.ts`** (new) — `runtimeWith(env, connectionConfig?, connectionId?)` shared test factory, replacing previously duplicated helpers in `index.test.ts` and `base-tools.test.ts`. Accepts an optional `connectionConfig` block so tests can populate service sub-blocks (kafka, flink, etc.) alongside the env, as needed for forward-compatible test writing during the 216–226 migration window.

## Why `getRequiredEnvVars()` was removed from `ToolHandler`

The interface represents the public contract that callers of a handler depend on. `getToolHandlersToRegister()` in `index.ts` was the only non-handler caller, and it now calls `enabledConnectionIds()` instead. With zero external callers remaining, leaving `getRequiredEnvVars()` on the interface would only be noise — and noise that carries `EnvVar` into the public surface, signalling that tool enablement is env-var-centric when it soon won't be. Moving it to `BaseToolHandler` as a protected implementation detail makes the intent clear: it is a shim hook, not a contract.

All existing handler implementations continue to satisfy the abstract requirement on `BaseToolHandler` unchanged — no handler files were touched.

## Why the disabled-tool log message changed

The old message named the specific missing env vars:
`"Tool X disabled due to missing environment variables: KAFKA_API_KEY, KAFKA_API_SECRET"`

`enabledConnectionIds()` returns a `string[]`, not a diagnostic payload, so the missing-var list is no longer available at the call site. More importantly, the old framing is wrong for the world this PR is building toward: once handlers migrate to typed predicates, a tool will be disabled because its required service block is absent from the connection config — not because an env var is missing. The new message:
`"Tool X disabled; no connections satisfy its requirements"`
reads correctly both today and after the migration is complete.

## Manual testing instructions

No functional changes — the same tools enable and disable as before. `npm run test` is the completeness check.

## Relationship to #173 and future domain migration PRs

This is the infrastructure PR (PR 0) described in the incremental migration plan in issue #173. It unblocks the eleven domain migration issues:

| Issue | Domain       |
| ----- | ------------ |
| #216  | billing      |
| #217  | catalog      |
| #218  | clusters     |
| #219  | connect      |
| #220  | environments |
| #221  | flink        |
| #222  | kafka        |
| #223  | metrics      |
| #224  | schema       |
| #225  | search       |
| #226  | tableflow    |

Each of those PRs will, for its domain's handlers:

1. Override `enabledConnectionIds()` with the appropriate one-liner predicate from `connection-predicates.ts`, e.g. `return connectionIdsWhere(runtime.config.connections, hasKafka)`.
2. Delete the `getRequiredEnvVars()` implementation from each handler.
3. Write an `enabledConnectionIds()` test for each handler (see test plan below) — none exist yet outside of `list-topics-handler.test.ts`, which serves as the template.

When the last domain PR lands, `getRequiredEnvVars()`, `runtime.env`, and the `FLINK_REQUIRED_ENV_VARS` / `CCLOUD_CONTROL_PLANE_REQUIRED_ENV_VARS` / etc. constants in `env-schema.ts` are deleted together in the cutover PR, redeclaring `enabledConnectionIds()` as `abstract` in `BaseToolHandler` to enforce that no handler can fall back to a shim that no longer exists.

## Test writing plan for domain migration PRs

The `list-topics-handler.test.ts` `enabledConnectionIds()` block is the template. At the time of writing (shim era), it tests against env vars:

```typescript
it("should return the connection ID when BOOTSTRAP_SERVERS is set", () => {
  const runtime = runtimeWith(envFactory({ BOOTSTRAP_SERVERS: "broker:9092" }));
  expect(handler.enabledConnectionIds(runtime)).toEqual(["default"]);
});

it("should return an empty array when BOOTSTRAP_SERVERS is absent", () => {
  const runtime = runtimeWith(envFactory());
  expect(handler.enabledConnectionIds(runtime)).toEqual([]);
});
```

When the domain migration PR for kafka (#222) lands, these two tests are updated to test against the connection config instead — and crucially, no env vars are provided, proving the handler no longer consults `runtime.env`:

```typescript
it("should return the connection ID for a connection with a kafka block", () => {
  const runtime = runtimeWith(
    envFactory(), // no env vars needed
    { kafka: { bootstrap_servers: "broker:9092" } }, // predicate checks this
  );
  expect(handler.enabledConnectionIds(runtime)).toEqual(["default"]);
});

it("should return an empty array for a connection without a kafka block", () => {
  const runtime = runtimeWith(envFactory()); // no env vars, no kafka block
  expect(handler.enabledConnectionIds(runtime)).toEqual([]);
});
```

The `runtimeWith()` factory in `tests/factories/runtime.ts` already accepts the optional `connectionConfig` second argument to support this pattern.

Closes #227.

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes

### DIFF
--- a/src/confluent/tools/base-tools.test.ts
+++ b/src/confluent/tools/base-tools.test.ts
@@ -1,0 +1,64 @@
+import { type CallToolResult } from "@src/confluent/schema.js";
+import {
+  BaseToolHandler,
+  READ_ONLY,
+  type ToolConfig,
+} from "@src/confluent/tools/base-tools.js";
+import { ToolName } from "@src/confluent/tools/tool-name.js";
+import type { EnvVar } from "@src/env-schema.js";
+import { envFactory } from "@tests/factories/env.js";
+import { runtimeWith } from "@tests/factories/runtime.js";
+import { describe, expect, it } from "vitest";
+
+class StubHandler extends BaseToolHandler {
+  constructor(private readonly vars: readonly EnvVar[]) {
+    super();
+  }
+
+  getRequiredEnvVars(): readonly EnvVar[] {
+    return this.vars;
+  }
+
+  getToolConfig(): ToolConfig {
+    return {
+      name: ToolName.LIST_TOPICS,
+      description: "stub",
+      inputSchema: {},
+      annotations: READ_ONLY,
+    };
+  }
+
+  handle(): CallToolResult {
+    return this.createResponse("stub");
+  }
+}
+
+describe("BaseToolHandler", () => {
+  describe("enabledConnectionIds()", () => {
+    it("should return the connection ID when all required env vars are present", () => {
+      const handler = new StubHandler(["KAFKA_API_KEY", "KAFKA_API_SECRET"]);
+      const runtime = runtimeWith(
+        envFactory({ KAFKA_API_KEY: "key", KAFKA_API_SECRET: "secret" }),
+      );
+      expect(handler.enabledConnectionIds(runtime)).toEqual(["default"]);
+    });
+
+    it("should return an empty array when a required env var is missing", () => {
+      const handler = new StubHandler(["KAFKA_API_KEY", "KAFKA_API_SECRET"]);
+      const runtime = runtimeWith(envFactory({ KAFKA_API_KEY: "key" }));
+      expect(handler.enabledConnectionIds(runtime)).toEqual([]);
+    });
+
+    it("should return an empty array when all required env vars are missing", () => {
+      const handler = new StubHandler(["KAFKA_API_KEY", "KAFKA_API_SECRET"]);
+      const runtime = runtimeWith(envFactory());
+      expect(handler.enabledConnectionIds(runtime)).toEqual([]);
+    });
+
+    it("should return the connection ID when getRequiredEnvVars returns an empty array", () => {
+      const handler = new StubHandler([]);
+      const runtime = runtimeWith(envFactory());
+      expect(handler.enabledConnectionIds(runtime)).toEqual(["default"]);
+    });
+  });
+});
--- a/src/confluent/tools/connection-predicates.ts
+++ b/src/confluent/tools/connection-predicates.ts
@@ -31,7 +31,7 @@ export function hasTableflow(conn: ConnectionConfig): boolean {
 }
 
 export function connectionIdsWhere(
-  connections: Record<string, ConnectionConfig>,
+  connections: Readonly<Record<string, ConnectionConfig>>,
   predicate: ConnectionPredicate,
 ): string[] {
   return Object.entries(connections)
--- a/src/confluent/tools/connection-predicates.ts
+++ b/src/confluent/tools/connection-predicates.ts
@@ -1,0 +1,40 @@
+import type { ConnectionConfig } from "@src/config/models.js";
+
+export type ConnectionPredicate = (conn: ConnectionConfig) => boolean;
+
+export function hasKafka(conn: ConnectionConfig): boolean {
+  return conn.kafka !== undefined;
+}
+
+export function hasKafkaRest(conn: ConnectionConfig): boolean {
+  return conn.kafka?.rest_endpoint !== undefined;
+}
+
+export function hasSchemaRegistry(conn: ConnectionConfig): boolean {
+  return conn.schema_registry !== undefined;
+}
+
+export function hasConfluentCloud(conn: ConnectionConfig): boolean {
+  return conn.confluent_cloud !== undefined;
+}
+
+export function hasFlink(conn: ConnectionConfig): boolean {
+  return conn.flink !== undefined;
+}
+
+export function hasTelemetry(conn: ConnectionConfig): boolean {
+  return conn.telemetry !== undefined;
+}
+
+export function hasTableflow(conn: ConnectionConfig): boolean {
+  return conn.tableflow !== undefined;
+}
+
+export function connectionIdsWhere(
+  connections: Record<string, ConnectionConfig>,
+  predicate: ConnectionPredicate,
+): string[] {
+  return Object.entries(connections)
+    .filter(([, conn]) => predicate(conn))
+    .map(([id]) => id);
+}
--- a/src/confluent/tools/handlers/kafka/list-topics-handler.test.ts
+++ b/src/confluent/tools/handlers/kafka/list-topics-handler.test.ts
@@ -1,6 +1,8 @@
 import { KafkaJS } from "@confluentinc/kafka-javascript";
 import { DefaultClientManager } from "@src/confluent/client-manager.js";
 import { ListTopicsHandler } from "@src/confluent/tools/handlers/kafka/list-topics-handler.js";
+import { envFactory } from "@tests/factories/env.js";
+import { runtimeWith } from "@tests/factories/runtime.js";
 import {
   createMockAdmin,
   createMockInstance,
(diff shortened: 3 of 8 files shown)


src/confluent/tools/connection-predicates.ts
```diff
@@ -0,0 +1,40 @@
...
+export function hasTelemetry(conn: ConnectionConfig): boolean {
+  return conn.telemetry !== undefined;
+}
+
+export function hasTableflow(conn: ConnectionConfig): boolean {
+  return conn.tableflow !== undefined;
+}
+
+export function connectionIdsWhere(
+  connections: Record<string, ConnectionConfig>,
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/confluent/tools/handlers/kafka/list-topics-handler.test.ts
```diff
@@ -28,6 +30,20 @@ describe("list-topics-handler.ts", () => {
       });
     });
 
+    describe("enabledConnectionIds()", () => {
+      it("should return the connection ID when BOOTSTRAP_SERVERS is set", () => {
+        const runtime = runtimeWith(
+          envFactory({ BOOTSTRAP_SERVERS: "broker:9092" }),
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/confluent/tools/base-tools.test.ts
shouples, line 1: Overall this makes sense, but on initial read it seems like it' testing `kafka` tool specifics instead of more generic logic. It might help to add a comment around the StubHandler constructor explaining the env var setting pattern, and use something like `name: "TEST TOOL" as ToolName` and generalized env var strings instead of `KAFKA_API_*`.

src/index.test.ts
```diff
@@ -28,7 +25,12 @@ import {
 
 function fakeHandler(requiredVars: EnvVar[], isCloudOnly = false): ToolHandler {
   return {
-    getRequiredEnvVars: () => requiredVars,
+    enabledConnectionIds: (runtime) => {
+      const missing = requiredVars.filter((v) => !runtime.env[v]);
+      return missing.length === 0
+        ? Object.keys(runtime.config.connections)
+        : [];
```
shouples, line 32: Could we repurpose `StubHandler` to be used here as well? Looks like some test helper overlap

shouples (approved):
LGTM, just some cleanup/consolidation suggestions to help future test writing

src/confluent/tools/base-tools.test.ts
jlrobins: class Environment only has definite in-use-already env var strings ;-( . But with some luck, the entirety of testing the shim implementation of enabledConnectionIds() (and therefore the env var interaction) will evaporate Real Soon with the completion of the entire sub-epic.

src/index.test.ts
jlrobins: Will look into in the next ticket, thanks!

jlrobins: #259 forked off for a separate, focused refactoring. Thanks!

src/confluent/tools/handlers/kafka/list-topics-handler.test.ts
```diff
@@ -28,6 +30,21 @@ describe("list-topics-handler.ts", () => {
       });
```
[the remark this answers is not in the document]
jlrobins: The intent is for it to fail then, at which time the test gets updated. TDD!

Issue #228, opened 2026-04-22
Cutover: redeclare enabledConnectionIds() abstract, delete getRequiredEnvVars() and constants


jlrobins:
The final PR for #173, after all domain migration children (#216–#226) are complete.

- Redeclare `enabledConnectionIds()` as `abstract` in `BaseToolHandler`, removing the shim. The compiler now enforces that every handler has a real implementation.
- Delete `getRequiredEnvVars()` and `isConfluentCloudOnly()` from `BaseToolHandler` and all handler implementations.
- Delete `TELEMETRY_REQUIRED_ENV_VARS`, `CCLOUD_CONTROL_PLANE_REQUIRED_ENV_VARS`, `CCLOUD_SCHEMA_REGISTRY_REQUIRED_ENV_VARS`, `KAFKA_REST_REQUIRED_ENV_VARS`, and `FLINK_REQUIRED_ENV_VARS` from `env-schema.ts` if no other usages remain.

PR #270, opened 2026-04-29
Add read-only connector tools and surface lcc-id

ashwinpankaj:
## Summary

Drops the `read-connector` tool and adds four narrower READ_ONLY Cloud Connect tools that, between them, cover the old tool's content and expose three views it never returned. Also surfaces the connector resource ID (`lcc-...`) on the new status tool so downstream "by connectorId" tools (metrics, logs) don't need a separate lookup.

**Why drop `read-connector`:** it returned `{name, config, tasks: [id-refs], type}` from `GET /connect/v1/.../connectors/{name}`. Its content is now covered by the more granular tools: the config map by `get-connector-config`, the per-task identifiers (and the connector type, derivable from `connector.class`) by `get-connector-tasks`. Callers that previously read the whole blob to pick one field can now ask for just that field.

**New tools:**
- `get-connector-config` — flat config map (`GET /connectors/{name}/config`)
- `get-connector-offsets` — per-task offsets (`GET /connectors/{name}/offsets`) — **new view, not in `read-connector`**
- `get-connector-status` — state + per-task health (`GET /connectors/{name}/status`), plus `lccId` projection — **new view, not in `read-connector`**
- `get-connector-tasks` — per-task configs (`GET /connectors/{name}/tasks`) — **new view; `read-connector` only returned task id-refs**

**`lccId` surfacing:** `get-connector-status` forwards `?expand=id` and projects the returned `id.id` as a top-level `lccId`. The OpenAPI schema doesn't declare the `id` augmentation, so the handler casts at the response boundary; the typed `Connector` schema stays a strict subset. `openapi.json` declares the optional `expand` query param on `/connectors/{name}/status`; types regenerated.

**Convention compliance** (post-rebase against #474/#515):
- All four handlers extend `ConnectToolHandler` (the domain base providing `predicate = hasConfluentCloud` and `category = ToolCategory.Connect`).
- Shared Zod input schema lifted to `connectorByNameArguments` in `connect-tool-handler.ts` — all four handlers reference it.
- Each new tool has a row in `EXPECTED_PREDICATES` in `tool-registry.test.ts` and a classification in `EXPECTED_OAUTH_DISABLED` in `src/index.test.ts`.

**Integration coverage:** `get-connector-status-handler.integration.test.ts` provisions a Datagen connector and asserts the `lccId` projection works end-to-end. Compensates for the deleted `read-connectors-handler.integration.test.ts`.

**Docs:**
- `CHANGELOG.md` Unreleased: 4 new tools under Added (framed as new views; three expose data `read-connector` never returned), `read-connector` under Removed with rationale.
- `README.md`: Connectors table row and alphabetical tool list updated; no remaining `read-connector` mentions outside the CHANGELOG.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test:unit` passes (1646 tests; colocated tests replaced with the modern `getMockedClientManager` + `assertHandleCase` pattern)
- [ ] Reviewer: confirm `openapi.json` diff is just the `expand` query-param block on `/connectors/{name}/status`
- [ ] Reviewer: confirm no `read-connector` mentions remain outside the `CHANGELOG.md` removal entry (`grep -rn read-connector` should match only the CHANGELOG line)
- [ ] Reviewer: spot-check one of the four new handlers against `list-connectors-handler.ts` for shape consistency
- [ ] Manual: invoke `get-connector-status` against a real Confluent Cloud connector and verify the response contains `"lccId":"lcc-..."`
- [ ] Manual: invoke `get-connector-config`, `get-connector-offsets`, `get-connector-tasks` against the same connector and verify each returns its own slice

Assisted-by: Claude Code


shouples (dismissed):
Thanks for these additions! A few things need fixing before this can merge.

- All new handlers still use the pre-YAML-migration pattern. The project moved to connection-predicate-based enablement in #228, and `src/confluent/tools/tool-registry.test.ts` explicitly fails any handler that still implements the old interface. Use `read-connectors-handler.ts` as the reference shape.

Quick reference - for each handler, swap:

- `getRequiredEnvVars()` -> use `enabledConnectionIds(runtime: ServerRuntime): string[]`
- `isConfluentCloudOnly(): boolean { return true; }` -> use `connectionIdsWhere(runtime.config.connections, hasConfluentCloud)`
- imports from `@src/env-schema.js` -> use `@src/confluent/tools/connection-predicates.js` + `@src/server-runtime.js`
- pre-#228 handler patterns -> use `read-connectors-handler.ts` as the reference shape

Also: the colocated tests should replace the `getRequiredEnvVars()` / `isConfluentCloudOnly()` describe blocks with `enabledConnectionIds()` coverage against `confluentCloudRuntime()` and `bareRuntime()` fixtures from `@tests/factories/runtime.js`.

To speed up the migration: open the handler files in Claude Code (or Copilot) and ask it to apply the #228 pattern - `.claude/rules/tool-handlers.md` and `.claude/rules/unit-tests.md` auto-load when editing files under `src/confluent/tools/**` and will walk through the exact changes needed.

jlrobins (dismissed):
Hi Ashwin, home stretch!

1. Needs to be updated from main, then new handler datamember `category` applied. See PR-474 for more details. You might desire to make an intermediate base class for all the connector handlers to define this as well as other commonalities between these (the Zod parameter shapes look Highly Similar and could also be centralized). 
2. Need to have integration test coverage over at least one of the new / replacement tools, otherwise is a net loss in integration test coverage.
3. The PR description has drifted from the content of the PR. Please revise to be self-consistent with the changes. Also ... citing the _reason_ why the read-connector tool should be replaced with these four would be great for posterity.
4. Needs CHANGELOG.md entries in the `Unreleased` section describing the dropped tool and the new added tools. Refer to how the 1.3.0 section is structured for end-user-visible additions and removals
5. README and CONFIGURATION documentation needs updating --- still mentions `read-connector`, needs to mention the new ones.  Grepping for `read-connector` should only end up finding the CHANGELOG entry describing its removal in favor of these new tools.

Let's get this PR clean and in, then the others should then be simple application of the same sort of changes. Thanks!

bcscc (approved):
LGTM from click-testing, thanks!

bcscc (approved):
Thanks! Ran integrations test locally and passed!

PR #271, opened 2026-04-29
Add connector lifecycle tools (pause/resume/restart/update-config)

ashwinpankaj:
## Summary

Adds four CREATE_UPDATE-annotated lifecycle tools for managing Cloud connectors:

- **pause-connector** — `PUT /connectors/{name}/pause`. Idempotent.
- **resume-connector** — `PUT /connectors/{name}/resume`. Idempotent.
- **restart-connector** — `POST /connectors/{name}/restart`. Asynchronous; the connector and its tasks are restarted as a single unit.
- **update-connector-config** — `PUT /connectors/{name}/config`. **Full-replace semantics: omitted keys are removed**, and the supplied map must include every key the connector's class requires (CCloud rejects otherwise).

All four extend `ConnectToolHandler` and reuse the `connectorByNameArguments` Zod schema introduced in #270; the `restart` tool reuses it verbatim, `update-connector-config` extends it with `connectorConfig: z.record(z.string(), z.string())`. They inherit `category = ToolCategory.Connect` and the `hasConfluentCloud` predicate from the base, so no per-tool predicate wiring is needed.

The prefix-convention test in `tool-registry.test.ts` gains `pause`/`resume`/`restart` in the `createUpdatePrefixes` set so the new tools satisfy convention enforcement.

## Review feedback addressed

- **Migrated to post-#228 patterns** (shouples): handlers no longer carry `getRequiredEnvVars()` / `isConfluentCloudOnly()` — predicate gating happens via the `ConnectToolHandler` base class's `predicate = hasConfluentCloud`.
- **Dropped the `as any` casts** (Copilot):
  - `restart-connector-handler.ts` no longer needs a cast. The original PR borrowed `includeTasks` / `onlyFailed` query flags from Apache Kafka's Connect REST API, but **CCloud's managed Connect does not expose them** (the upstream `openapi.json` declares no query params on `restartConnectv1Connector`). They've been removed from the tool's input schema. The handler is now a straight typed POST.
  - `update-connector-config-handler.ts` uses a narrower `as unknown as UpdateConfigBody` cast derived from the typed endpoint, scoped to the body only so path params stay fully type-checked. The cast acknowledges that the generated body type marks `connector.class` / `name` / `kafka.api.key` / `kafka.api.secret` as required while the handler accepts a runtime `Record<string, string>`.
- **Misleading comment fixed** (Copilot / shouples): the "validates the merged config server-side" comment on `update-connector-config-handler.ts` was inaccurate. The endpoint uses **replacement** semantics; comment now reflects that.

## Integration coverage

Adds `connector-lifecycle.integration.test.ts` — provisions a Datagen Source connector once and exercises all four lifecycle tools against it across every active transport (stdio / http / sse). Also adds `waitForConnectorRunnable` to the shared `tests/harness/connect.ts` so restart doesn't race the connector's PROVISIONING state.

## Docs

- `CHANGELOG.md`: Unreleased entry describing the four tools, with explicit notes on CCloud's lack of `includeTasks` / `onlyFailed` on restart and update-config's full-replace semantics.
- `README.md`: connector tools table row updated; alphabetical `--list-tools` cheatsheet picks up the four new tool descriptions.
- `CONFIGURATION.md`: connector predicate-block matrix updated. Also drops a stale `read-connector` reference left behind when #270 retired that tool — `grep read-connector` now matches only the CHANGELOG removal entry.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test:unit` passes (1760 tests)
- [x] `npm run test:integration -- src/confluent/tools/handlers/connect/connector-lifecycle.integration.test.ts` passes (15 tests; pause → resume → restart → update-config × stdio/http/sse against a real CCloud connector)
- [ ] Reviewer: confirm the four handlers extend `ConnectToolHandler` and reuse `connectorByNameArguments` (no per-tool Zod-arg duplication)
- [ ] Reviewer: confirm `update-connector-config-handler.ts`'s body cast is scoped to `body:` only (path params stay typed) and the comment matches PUT replacement semantics
- [ ] Reviewer: confirm `restart-connector-handler.ts` no longer takes `includeTasks` / `onlyFailed` (CCloud doesn't expose them)

Assisted-by: Claude Code


shouples (dismissed):
Thanks for adding these - similar to #270, each new handler in this PR implements `getRequiredEnvVars()` and `isConfluentCloudOnly()` (imported from `@src/env-schema.js`). The test at `src/confluent/tools/tool-registry.test.ts` asserts those methods are absent from every registered handler, so this PR will fail that test.

For reference:
- Use `enabledConnectionIds(runtime: ServerRuntime): string[]`, not `getRequiredEnvVars()`
- Use `connectionIdsWhere(runtime.config.connections, hasConfluentCloud)` as the body, not `isConfluentCloudOnly(): boolean`
- Import from `@src/confluent/tools/connection-predicates.js` + `@src/server-runtime.js`, not from `@src/env-schema.js`
- Model on `src/confluent/tools/handlers/connect/read-connectors-handler.ts`, not pre-#228 patterns

One small comment fix too: `update-connector-config-handler.ts` line ~66 says the API "validates the merged config server-side" but the actual semantics are full-replace - omitted keys are removed. Worth aligning with the tool description.

If you're using Claude Code or Copilot, run it over the handler files directly - `.claude/rules/tool-handlers.md` auto-loads when editing `src/confluent/tools/**/*.ts` and walks through the migration pattern step by step.

src/confluent/tools/handlers/connect/connector-lifecycle.integration.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        const result = await server.client.callTool({
+          name: ToolName.RESTART_CONNECTOR,
+          arguments: { connectorName },
+        });
+        expect(textContent(result)).toBe(
+          `Restart requested for connector ${connectorName}.`,
+        );
+      });
+
+      it("should replace the connector config via update-connector-config", async () => {
```
bcscc, line 108: I'm running into a failure for this test, see:
```
 FAIL   integration  src/confluent/tools/handlers/connect/connector-lifecycle.integration.test.ts > connector-lifecycle > via stdio transport > should restart the connector
AssertionError: expected 'Failed to restart connector int-conne…' to be 'Restart requested for connector int-c…' // Object.is equality

Expected: "Restart requested for connector int-connect-lifecycle-1780335274450-qr684q."
Received: "Failed to restart connector int-connect-lifecycle-1780335274450-qr684q: {"error":{"code":400,"message":"Operation not supported for connector in PAUSED state"}}"
```

Seems like an async-state race, the lifecycle ops are async on CCloud's side, so the success message just confirms the request was accepted, not that the state transitioned. restart fires before resume propagates, so the connector's still PAUSED when CCloud 400s. Maybe this needs a poll for RUNNING between resume and restart?

src/confluent/tools/handlers/connect/pause-connector-handler.test.ts
```diff
@@ -0,0 +1,132 @@
...
+            ),
+            args,
+            outcome,
+            clientManager,
+          });
+
+          if (typeof outcome === "object" && "resolves" in outcome) {
+            expect(cloudRest.PUT).toHaveBeenCalledOnce();
+            expect(cloudRest.PUT).toHaveBeenCalledWith(
+              expect.any(String),
```
bcscc: `expect.any(String)` doesn't pin the path, these three handlers differ only by suffix (`/pause//resume//restart`), so pause passes even if it hits `/resume`. Sibling `get-connector-config-handler.test.ts:110` pins it with `stringContaining("/config")`. Use a literal path (or `stringContaining("/pause")`)

src/confluent/tools/handlers/connect/restart-connector-handler.test.ts
```diff
@@ -0,0 +1,132 @@
...
+            ),
+            args,
+            outcome,
+            clientManager,
+          });
+
+          if (typeof outcome === "object" && "resolves" in outcome) {
+            expect(cloudRest.POST).toHaveBeenCalledOnce();
+            expect(cloudRest.POST).toHaveBeenCalledWith(
+              expect.any(String),
```
bcscc: Same here, if possible pin expect path.

src/confluent/tools/handlers/connect/resume-connector-handler.test.ts
```diff
@@ -0,0 +1,132 @@
...
+            ),
+            args,
+            outcome,
+            clientManager,
+          });
+
+          if (typeof outcome === "object" && "resolves" in outcome) {
+            expect(cloudRest.PUT).toHaveBeenCalledOnce();
+            expect(cloudRest.PUT).toHaveBeenCalledWith(
+              expect.any(String),
```
bcscc: Same here

bcscc (commented):
Thanks @ashwinpankaj !! Tools handlers look great, just running into 1 integration test failure + a couple nits for a unit test.

bcscc (approved):
LGTM Thanks!

PR #273, opened 2026-04-29
Add connector error-diagnostics and logs tools

ashwinpankaj:
## Summary

Three READ_ONLY Cloud-only diagnostic tools for connector troubleshooting, all extending `ConnectToolHandler` (predicate `hasConfluentCloud`, `category = ToolCategory.Connect`):

### Error diagnostics

- **get-connector-error-summary** — projects Confluent Cloud's `/status` payload's Cloud-only diagnostic extras (`error_summary`, `validation_errors`, `errors_from_trace`, `override_message`, `error_details`, `is_csfle_error`, `plugin_lifecycle`, per-task trace heads) into a compact summary. Healthy connectors short-circuit to a one-liner. The OpenAPI schema doesn't declare these extras, so we cast at the response boundary.

- **get-connector-error-recommendations** — thin projection over `GET /connectors/{name}/error-recommendations`. Returns Cloud-generated remediation strings plus `eventId` for support cross-reference. Cloud's 400 "no stack trace yet" response is treated as a friendly one-liner rather than a tool error.

Adds `/connectors/{name}/error-recommendations` to `openapi.json` and surgically splices the new operation into `openapi-schema.d.ts` (only the new path + operation entries, not a full regen).

### Logs

- **get-connector-logs** — retrieves recent connector log entries from `api.logging.confluent.cloud` (separate host from `api.confluent.cloud`, not declared in `openapi.json`).

  **Auth flow** (two-step, mirrors confluentinc/cli's `pkg/auth/auth.go` + `pkg/ccloudv2/logging.go`):

  1. POST `{confluent.cloud}/api/access_tokens` with HTTP Basic auth (`CONFLUENT_CLOUD_API_KEY` + `_SECRET`) → short-lived data-plane bearer token. Cloud rejects Basic auth directly on the logging API with `401 'Credentials are required to access this resource'`.
  2. POST `{api.logging.confluent.cloud}/logs/v1/search?page_size=N` with `Authorization: Bearer <token>`.

  **Pagination**: optional `pageToken` arg forwarded as `?page_token=<value>`; `metadata.next` from the response surfaces as `nextPageToken` so the model can iterate.

  **Org ID resolution order**: `organizationId` arg → `conn.confluent_cloud.organization_id` (seeded from YAML or, on the legacy env-var path, from `CONFLUENT_CLOUD_ORG_ID`) → auto-fetch via `GET /org/v2/organizations` (uses `data[0].id`).

  Adds `CONFLUENT_CLOUD_ORG_ID` to `env-schema.configSchema` (effectively optional — the whole configSchema is `.partial()`) and plumbs it into `models.ts` (`confluent_cloud.organization_id`) and `env-config.ts` (`ENV_VAR_TO_ZPATH`).

## Review feedback addressed

- **Post-#228 patterns** (shouples): handlers extend `ConnectToolHandler` (which carries `predicate = hasConfluentCloud` and `category = ToolCategory.Connect`); no `getRequiredEnvVars()` / `isConfluentCloudOnly()`.
- **Reuse `connectorByNameArguments`** (the shared Zod schema introduced for the Connect domain in #270): all three handlers drop their local Zod arg duplicates. The logs handler extends the shared schema with the log-specific fields (`organizationId`, `connectorId`, `levels`, `startTime`, `endTime`, `pageSize`, `pageToken`).
- **Org-ID via runtime config** (shouples): handler reads from `conn.confluent_cloud.organization_id`, not from `process.env` at call time. The env var still feeds that field via env-config translation for the legacy env-var path.
- **`CONFLUENT_CLOUD_ORG_ID` is optional** (shouples + Copilot inline): the `configSchema` is `.partial()`, so all of its keys (this one included) are optional. The description in `env-schema.ts` now reflects that: optional; setting it skips the auto-resolution hop, otherwise the logs handler resolves the org via the Cloud API.

## Integration coverage

`connector-error-diagnostics.integration.test.ts` provisions two connectors in parallel and exercises all three tools end-to-end across every active transport:

- **Failed PostgresSource** (`connection.host = nonexistent.invalid.example.com`) — CCloud accepts the POST then fails at config validation. Exercises `error-summary`'s populated projection (`hasError:true`, `state:"FAILED"`, parsed `validationErrors`) and `error-recommendations`'s response surface.
- **Healthy Datagen Source** — exercises the logs token-exchange + `/logs/v1/search` round-trip against a connector that actually emits INFO-level lifecycle entries.

New helpers in `tests/harness/connect.ts`: `provisionFailedTestPostgresConnector`, `waitForConnectorFailed` (3-minute default timeout because runtime failures lag PROVISIONING by 30–90s).

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test:unit` passes (1809 tests)
- [x] `npm run test:integration -- src/confluent/tools/handlers/connect/connector-error-diagnostics.integration.test.ts` passes (12 tests: 4 ops × 3 transports against real CCloud)
- [ ] Reviewer: confirm `openapi.json` adds exactly the `/error-recommendations` path block; nothing else
- [ ] Reviewer: confirm `openapi-schema.d.ts` diff is exactly one new path entry + one new operation entry, not a full regen
- [ ] Reviewer: confirm `tool-name.ts` adds exactly 3 entries; `tool-registry.ts` adds 3 imports + 3 map entries
- [ ] Reviewer: spot-check `exchangeForDataPlaneToken()` in `get-connector-logs-handler.ts` against the cli reference flow

Assisted-by: Claude Code


shouples (dismissed):
Thanks for this - similar migrations are needed to #270 and #271 though:
- Use `enabledConnectionIds(runtime)`, not `getRequiredEnvVars()`
- Use `connectionIdsWhere(runtime.config.connections, hasConfluentCloud)`, not `isConfluentCloudOnly(): boolean`
- Use imports from `@src/confluent/tools/connection-predicates.js` + `@src/server-runtime.js`, not from `@src/env-schema.js`

One other thing around the CCloud org ID use:
- Use the resolved `confluent_cloud.organization_id` from the runtime connection config (via clientManager / connection), not `config.env.CONFLUENT_CLOUD_ORG_ID` re-read at call time
- The `organization_id` field added to `models.ts` is the right place - handlers should read from there, not bypass it via env
- `CONFLUENT_CLOUD_ORG_ID` in `env-schema.ts` needs `.optional()` unless it's truly needed for all `confluent_cloud` configs (which would then need additional tool handler updates)

src/confluent/tools/handlers/connect/get-connector-error-recommendations-handler.ts
```diff
@@ -0,0 +1,118 @@
...
+} from "@src/confluent/tools/handlers/connect/connect-tool-handler.js";
+import { ToolName } from "@src/confluent/tools/tool-name.js";
+import { ServerRuntime } from "@src/server-runtime.js";
+import { wrapAsPathBasedClient } from "openapi-fetch";
+
+interface RecommendationsProjection {
+  connectorName: string;
+  recommendations: string[];
+  eventId?: string;
+  engineError?: unknown;
```
bcscc:
```suggestion
```
This could be removed, unused member

src/config/env-config.ts
bcscc, line 317: The new `CONFLUENT_CLOUD_ORG_ID` is declare and mapped in `ENV_VAR_TO_ZPATH`, but it's never spread into the block here, so on the env-var path it's silently dropped. Include to match the YAML path and a covering test in env-config.test.ts

bcscc (approved):
Overall LGTM! `@connect` Integration tests are passing when ran locally, just 2 comments that I'm also happy to address as follow-ups.

bcscc:
Also please update the README for the new tools and add an entry to the change log!

bcscc (approved):
Thanks!! Great new tool additions!

PR #295, opened 2026-04-30
Feat/229 handle() gets runtime, not clientmanage passed now

### LINKED ISSUE #229: Pass `ServerRuntime` to `handle()` instead of ClientManager
## Background

`handle(clientManager, toolArguments, sessionId)` receives a bare `ClientManager` with no
connection context. Six handlers also reach outside `handle()` entirely, reading operational
defaults (`FLINK_ENV_ID`, `KAFKA_CLUSTER_ID`, etc.) from the global `env` singleton at call time.

`ServerRuntime` is already the single object that ties config to client managers. Passing it
into `handle()` instead of the bare `ClientManager` lets every handler derive its config values
from the connection it was constructed from, eliminating the global reads.

## Migration

This is a large mechanical change — every `handle()` call site and implementation needs updating. Approach:

1. Change `handle()` signature in `BaseToolHandler` and `ToolHandler` interface.
2. Update all handler implementations and call sites

jlrobins:
## Summary of Changes

Widen the contract between the MCP server and its tool handlers: `handle()` now receives the full `ServerRuntime` instead of a single `ClientManager`. This is a prerequisite for two upcoming capabilities:

- **Multi-connection support** — when multiple connections are configured, tools will gain an explicit invocation parameter indicating which connection to operate against; `runtime` gives handlers access to the full connection map needed to resolve that parameter to a client.
- **OAuth / Confluent Cloud unified connection** — a single CCloud-via-OAuth connection provides access to every resource the user owns across services. This will likely require a different client abstraction than `ClientManager`; having `runtime` in the signature gives handlers a place to receive it without another interface change.

The `const clientManager = runtime.clientManager;` shim construct in each handler preserves all existing handler bodies unchanged. It is the seam that will be removed as multi-connection and OAuth routing are implemented in follow-on tickets.

## Changes by file category

  1. **`src/confluent/tools/base-tools.ts`** — `ToolHandler` interface and `BaseToolHandler` abstract class: `handle(clientManager, ...)` → `handle(runtime, ...)`.

  2. **`src/index.ts`** — single call site updated to pass `runtime` instead of `runtime.clientManager`.

  3. **`tests/factories/runtime.ts`** — `runtimeWith()` gains an optional third argument to inject a pre-configured `Mocked<DefaultClientManager>`, so handler tests that call `handle()` directly can still control the mock client without constructing a full runtime from scratch.

  4. **`tests/server.ts` + `src/mcp/server.test.ts`** — in-process MCP test harness updated to pass a `ServerRuntime` rather than a bare `ClientManager`.

  5. **`tests/stubs/stub-handler.ts`** — `StubHandler.handle()` signature updated to match the new interface.

  6. **`.claude/rules/unit-tests.md`** — handler test pattern documentation updated to show the `runtimeWith(..., clientManager)` call shape.

  7. **50 concrete handler classes** (`handlers/billing`, `catalog`, `clusters`, `connect`, `environments`, `flink/**`, `kafka`, `metrics`, `schema`, `search`, `tableflow/**`) — purely mechanical: signature updated, `const clientManager = runtime.clientManager;` added as first body line. No logic changes.

## Smoke test suite

`tool-registry.test.ts` gains an exhaustive `handle()` smoke test that drives every registered handler with empty arguments against a fully-wired universal-proxy `ClientManager`, lifting overall line coverage from 37% to 50.19%. Each handler's expected outcome is declared in a `Partial<Record<ToolName, HandleSmokeCase>>` table — typed as a discriminated union of `{ resolves: string }` and `{ throws: string }` — with a parameterized completeness check that gives each missing tool its own failing test. Handler invocation is separated from assertions, and a precondition guard prevents empty-string `resolves` placeholders from masquerading as real assertions.

When a developer adds a new tool and forgets the table, a named test fails with the exact line to paste in (`[ToolName.FOO]: "TODO"`). Replacing the placeholder with `"TODO"` and re-running the tests drives the handler against the universal client and reports the correct entry to substitute — a "golden file"-style discovery workflow that guides the developer to a passing state without requiring them to read any documentation.

The universal client is a **two-proxy pair** rather than a single recursive proxy. `callableProxy` (function target) absorbs any method chain or call without throwing; every async call resolves to `responseProxy` (plain-object target), which models the `{ data, error }` shape returned by openapi-fetch: `error` is always `undefined` (i.e. "no API error"), and `data` returns the per-entry `responseData` value. This means handlers now advance past the `if (error)` guard and into their success-path code rather than immediately returning error messages. Two handlers — `list-clusters` and `list-available-metrics` — supply `{ data: [] }` as `responseData` and reach their genuine success branch with no error handling involved. The `responseData` field is optional and defaults to `{}`; remaining entries that stop at Zod validation or a guard check are left as-is for a future sitting.

> **Follow-on:** `stubClientGetters` and its two-proxy pair could be promoted out of the smoke-test `describe` block and into `tests/stubs/` or `tests/factories/`, where individual handler unit tests can import and reuse it. Supplying a realistic `responseData` shape eliminates the need to manually stub the full `pathBasedClient["/endpoint"].GET({params})` call chain in per-tool tests — replacing five lines of mock wiring with one.

---

_Miles to go before we sleep — significant design, planning, and implementation work remains before the single-connection shim construct can be removed from the handler bodies._

- No feature changes.
- Closes #229 

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes

### DIFF
--- a/src/confluent/tools/tool-registry.test.ts
+++ b/src/confluent/tools/tool-registry.test.ts
@@ -182,44 +182,66 @@ describe("tool-registry.ts", () => {
 
     /**
      * Wires every client getter on a fresh `Mocked<DefaultClientManager>` to a
-     * recursive Proxy: any property access or call on the returned client resolves
-     * to another instance of the same Proxy, so handler bodies never throw a
-     * TypeError before they reach real logic. A TypeError that escapes is therefore
-     * a genuine wiring failure — exactly what the smoke tests check for.
+     * two-proxy pair. `callableProxy` (function target) absorbs any method chain
+     * or call without throwing; every call resolves to `responseProxy`.
+     * `responseProxy` (plain-object target) models the `{ data, error }` shape
+     * returned by openapi-fetch: `error` is `undefined`, and `data` returns
+     * `responseData` — supply a minimal valid shape to push a handler past schema
+     * validation into its real success branch (defaults to `{}`).
+     *
+     * A TypeError that escapes the proxy pair is therefore a genuine wiring
+     * failure — exactly what the smoke tests check for.
      *
      * Returns `{ clientManager, clientGetters }`. `clientGetters` is the canonical
      * list of all getter mocks; callers use it to assert that at least one was
      * invoked when a handler resolves successfully.
      */
-    function stubClientGetters() {
-      const proxy = new Proxy(() => Promise.resolve(proxy), {
+    function stubClientGetters(responseData: unknown = {}) {
+      // Two-proxy setup: callableProxy (function target) handles method chains
+      // and calls; responseProxy (plain-object target) is what async calls
+      // resolve to, giving { data, error } destructuring a real object for `data`.
+      // responseData is returned for the `data` property — callers can supply a
+      // minimal valid shape to push a handler past validation into its success path.
+      let responseProxy: object = {};
+      const callableProxy = new Proxy((() => {}) as () => Promise<object>, {
         get: (_t, prop) => {
           if (prop === "then" || prop === "error") return undefined;
-          return () => Promise.resolve(proxy);
+          return callableProxy;
         },
-        apply: () => Promise.resolve(proxy),
+        apply: () => Promise.resolve(responseProxy),
+      });
+      responseProxy = new Proxy({} as object, {
+        get: (_t, prop) => {
+          if (prop === "then" || prop === "error") return undefined;
+          if (prop === "data") return responseData;
+          return callableProxy;
+        },
       });
       const clientManager = createMockInstance(DefaultClientManager);
-      clientManager.getAdminClient.mockResolvedValue(proxy as never);
-      clientManager.getProducer.mockResolvedValue(proxy as never);
-      clientManager.getConsumer.mockResolvedValue(proxy as never);
+      clientManager.getAdminClient.mockResolvedValue(callableProxy as never);
+      clientManager.getProducer.mockResolvedValue(callableProxy as never);
+      clientManager.getConsumer.mockResolvedValue(callableProxy as never);
       clientManager.getConfluentCloudFlinkRestClient.mockReturnValue(
-        proxy as never,
+        callableProxy as never,
       );
-      clientManager.getConfluentCloudRestClient.mockReturnValue(proxy as never);
+      clientManager.getConfluentCloudRestClient.mockReturnValue(
+        callableProxy as never,
+      );
       clientManager.getConfluentCloudTableflowRestClient.mockReturnValue(
-        proxy as never,
+        callableProxy as never,
       );
       clientManager.getConfluentCloudSchemaRegistryRestClient.mockReturnValue(
-        proxy as never,
+        callableProxy as never,
       );
       clientManager.getConfluentCloudKafkaRestClient.mockReturnValue(
-        proxy as never,
+        callableProxy as never,
       );
       clientManager.getConfluentCloudTelemetryRestClient.mockReturnValue(
-        proxy as never,
+        callableProxy as never,
       );
-      clientManager.getSchemaRegistryClient.mockReturnValue(proxy as never);
+      clientManager.getSchemaRegistryClient.mockReturnValue(
+        callableProxy as never,
+      );
       const clientGetters = [
         clientManager.getAdminClient,
         clientManager.getProducer,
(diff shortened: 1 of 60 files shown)


src/confluent/tools/tool-registry.test.ts
```diff
@@ -123,4 +131,298 @@ describe("tool-registry.ts", () => {
...
+     * Wires every client getter on a fresh `Mocked<DefaultClientManager>` to a
+     * recursive Proxy: any property access or call on the returned client resolves
+     * to another instance of the same Proxy, so handler bodies never throw a
+     * TypeError before they reach real logic. A TypeError that escapes is therefore
+     * a genuine wiring failure — exactly what the smoke tests check for.
+     *
+     * Returns `{ clientManager, clientGetters }`. `clientGetters` is the canonical
+     * list of all getter mocks; callers use it to assert that at least one was
+     * invoked when a handler resolves successfully.
+     */
```
shouples: Looks like this should go above `stubClientGetters`

src/confluent/tools/tool-registry.test.ts
```diff
@@ -123,4 +131,298 @@ describe("tool-registry.ts", () => {
...
+      // Kafka
+      [ToolName.LIST_TOPICS]: { resolves: "Kafka topics:" },
+      [ToolName.CREATE_TOPICS]: { throws: "ZodError" },
+      [ToolName.DELETE_TOPICS]: { throws: "ZodError" },
+      [ToolName.PRODUCE_MESSAGE]: { throws: "ZodError" },
+      [ToolName.CONSUME_MESSAGES]: { throws: "ZodError" },
+      [ToolName.ALTER_TOPIC_CONFIG]: { throws: "ZodError" },
+      [ToolName.GET_TOPIC_CONFIG]: { throws: "ZodError" },
+      // Schema Registry
+      [ToolName.LIST_SCHEMAS]: { resolves: "Failed to list schemas" },
```
shouples: This seems odd - it resolves a failure message?

src/confluent/tools/tool-registry.test.ts
```diff
@@ -123,4 +131,298 @@ describe("tool-registry.ts", () => {
...
+        resolves: "Failed to fetch environments",
+      },
+      [ToolName.READ_ENVIRONMENT]: { throws: "ZodError" },
+      // Clusters
+      [ToolName.LIST_CLUSTERS]: { resolves: "Failed to fetch clusters" },
+      // Tableflow
+      [ToolName.CREATE_TABLEFLOW_TOPIC]: { throws: "ZodError" },
+      [ToolName.LIST_TABLEFLOW_REGIONS]: {
+        resolves: "Failed to list Tableflow regions",
+      },
```
shouples: Same with these handful of `resolves: "Failed to ...` scenarios

src/confluent/tools/tool-registry.test.ts
```diff
@@ -123,4 +131,298 @@ describe("tool-registry.ts", () => {
...
+        }
+
+        // Placeholder sentinel: run completed — report what to replace it with
+        if (outcome === "TODO") {
+          const discovered =
+            thrown === undefined
+              ? `{ resolves: "${result!.content
+                  .map((c) => ("text" in c ? c.text : ""))
+                  .join("")
+                  .slice(0, 60)}" }`
```
shouples, line 440: Noting this section since my upcoming branch has a `textContent` parser util that would be good to add here (https://github.com/confluentinc/mcp-confluent/issues/296)

shouples (dismissed):
LGTM, no blockers but a couple of questions that can be addressed in small follow-up PRs

src/confluent/tools/tool-registry.test.ts
```diff
@@ -123,4 +131,298 @@ describe("tool-registry.ts", () => {
...
+      [ToolName.ADD_TAGS_TO_TOPIC]: { throws: "ZodError" },
+      [ToolName.LIST_TAGS]: { resolves: "Failed to list tags" },
+      // Search
+      [ToolName.SEARCH_TOPICS_BY_TAG]: {
+        resolves: "Failed to search for topics by tag",
+      },
+      [ToolName.SEARCH_TOPICS_BY_NAME]: { throws: "ZodError" },
+      // Environments
+      [ToolName.LIST_ENVIRONMENTS]: {
+        resolves: "Failed to fetch environments",
```
shouples: Actually something seems off with these `resolves: "Failed to..."` assertions. For this `Failed to fetch environments`, we'd only return that if there was an error response:
https://github.com/confluentinc/mcp-confluent/blob/0b824bf9ba997813f6cf07d40080cfbdfb730530/src/confluent/tools/handlers/environments/list-environments-handler.ts#L90
I'd expect to see something more like this:
https://github.com/confluentinc/mcp-confluent/blob/0b824bf9ba997813f6cf07d40080cfbdfb730530/src/confluent/tools/handlers/environments/list-environments-handler.ts#L131

src/confluent/tools/tool-registry.test.ts
```diff
@@ -123,4 +131,298 @@ describe("tool-registry.ts", () => {
...
+      if (thrown instanceof Error) return thrown.message;
+      throw new Error(
+        `Wacky -- ${name}: handler threw a non-Error value: ${JSON.stringify(thrown)}`,
+      );
+    }
+
+    function stubClientGetters() {
+      const proxy = new Proxy(() => Promise.resolve(proxy), {
+        get: (_t, prop) => {
+          if (prop === "then") return undefined;
```
shouples: I *think* this line may be the problem - the `error` is property is being turned into `Promise.resolve(proxy)` which is making these truthy:
```ts
const { data: response, error } = await pathBasedClient[ ... ] // path+method call here
if (error) {
   logger.error({ error }, "API Error");
   return this.createResponse(
      `Failed to fetch ... `
   // etc
}
```

By checking `error` here in the proxy `get`, I see responses closer to what I'd expect when re-running the unit tests:

Image

However, this may require some more stubbing/mocking for some of the post-response data handling.

shouples (changes requested):

src/confluent/tools/tool-registry.test.ts
jlrobins: Oh someone's going to get humiliated for injecting a new method between an existing one and its doc-comment.

src/confluent/tools/tool-registry.test.ts
jlrobins: Same answer.

src/confluent/tools/tool-registry.test.ts
jlrobins: Fixed using a two-level proxy and sometimes supplying the API call response structure.

src/confluent/tools/tool-registry.test.ts
jlrobins: Fixed with two-level proxy.

src/confluent/tools/tool-registry.test.ts
jlrobins: Applied and then more, thanks, good analysis!

src/confluent/tools/tool-registry.test.ts
```diff
@@ -182,44 +182,66 @@ describe("tool-registry.ts", () => {
...
      * invoked when a handler resolves successfully.
      */
-    function stubClientGetters() {
-      const proxy = new Proxy(() => Promise.resolve(proxy), {
+    function stubClientGetters(responseData: unknown = {}) {
+      // Two-proxy setup: callableProxy (function target) handles method chains
+      // and calls; responseProxy (plain-object target) is what async calls
+      // resolve to, giving { data, error } destructuring a real object for `data`.
+      // responseData is returned for the `data` property — callers can supply a
+      // minimal valid shape to push a handler past validation into its success path.
```
shouples: Nice!

shouples (approved):
👏

PR #301, opened 2026-05-01
Add OAuthClientManager and wire into ServerRuntime

### LINKED ISSUE #298: Add OAuthClientManager and wire into ServerRuntime
- New OAuthClientManager extends BaseClientManager: wires bearer middleware on every REST client at construction time (not lazy), pulls cloud REST endpoint from getCloudRestUrlForEnv(env),  no native Kafka client.
- When OAuth connection detected, construct OAuthClientManager instead of DirectClientManager.
  - Tools gated by hasConfluentCloud become OAuth-eligible automatically (control-plane subset of the 43 REST tools).
  - Per-surface manual verification against stag for each cloud control-plane tool.

bcscc:
## Summary of Changes

- Add `OAuthClientManager extends BaseClientManager implements ClientManager` (new file `src/confluent/oauth-client-manager.ts`). Wires every REST surface to bearer tokens supplied by the existing `OAuthHolder` — control-plane (cloud / tableflow / telemetry) reads `getControlPlaneToken()`; data-plane (flink / SR-REST / kafka-REST) reads `getDataPlaneToken()`. Cloud REST URL is auto-derived from the Auth0 environment via the new `getCloudRestUrlForEnv` helper. The constructor eagerly warms the cloud REST client so a bad endpoint or middleware wiring fails fast at startup rather than at first tool call.
  - Native Kafka methods (`getKafkaClient` / `getAdminClient` / `getProducer` / `getConsumer`) throw with an OAuth-specific error. They're gated off by the `hasKafka` predicate (the OAuth synth connection has no `kafka` block) so the throws are defensive.
  - `getSchemaRegistryClient()` is overridden to throw a clearer "use the REST client" message under OAuth, so callers don't get the inherited "configure SCHEMA_REGISTRY_API_KEY" message that's misleading in an OAuth flow.
- Wire `OAuthClientManager` into `ServerRuntime.fromConfig`: when an OAuth connection is detected (`config.getCCloudOAuth()` returns a value), every connection's manager is the bearer-auth-backed `OAuthClientManager` sharing the single holder; otherwise fall back to the per-connection direct factory.
- Extend `Auth0Config` with a `cloudRestUrl` field (devel/stag/prod). `getCloudRestUrlForEnv(env)` exported from `auth0-config.ts`.

### Tests

- New `src/confluent/oauth-client-manager.test.ts` — cloud REST URL derivation per env, Tableflow URL reuse, data-plane endpoints stay undefined, eager cloud-client materialization, native Kafka methods throw, SR SDK getter throws OAuth-specific error, `disconnect()` resolves cleanly.
- Extended `src/server-runtime.test.ts` — asserts `runtime.clientManagers[id]` is an `OAuthClientManager` when `ccloudOAuth` is configured.
- Extended `src/confluent/oauth/auth0-config.test.ts` — covers the new `getCloudRestUrlForEnv` helper across all three envs.

### Feature Branch Stack (bcscc/oauth-298-oauth-client-manager)

- #299
  - #301 **(this)**

Fixes #298

- `list-billing-costs, list-clusters, list-environments, read-environment, list-connectors, read-connector, delete-connector` now all work when startup using `--oauth`

- **Known limitation, deferred**: when `--oauth --oauth-env=stag` is set together with env vars like `BOOTSTRAP_SERVERS` / `SCHEMA_REGISTRY_*` / `TELEMETRY_*`, those blocks land on the synthesized connection and predicates (`hasKafkaBootstrap`, `hasSchemaRegistry`, `hasTelemetry`) will enable tools whose runtime calls then fail (native Kafka throws, SR SDK throws OAuth-specific error). The fundamental ambiguity (auth source vs. block presence) is what the connection-shape migration resolves — out of scope here.
- **Telemetry endpoint stays unset under OAuth**, intentionally: predicate gating prevents telemetry tools from being enabled under OAuth (no `telemetry` block in the synth connection). Once a future layer lets users supply a telemetry endpoint via YAML for OAuth connections, it'll flow through naturally.
- Four pre-existing observations Copilot raised on this PR were tracked back to behavior that pre-dates the refactor (Lazy.get() not caching, Tableflow URL hardwired, `extra_properties` precedence, `as any` cast on `kafkaJS` config). Those are worth a follow-up cleanup ticket but kept out of this PR to maintain bit-identical direct-connection behavior and keep #298 narrowly focused on OAuth wiring.

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing

### DIFF
--- a/src/confluent/oauth-client-manager.test.ts
+++ b/src/confluent/oauth-client-manager.test.ts
@@ -1,0 +1,119 @@
+import { BaseClientManager } from "@src/confluent/base-client-manager.js";
+import { OAuthClientManager } from "@src/confluent/oauth-client-manager.js";
+import { OAuthHolder } from "@src/confluent/oauth/oauth-holder.js";
+import { describe, expect, it, vi } from "vitest";
+
+function fakeOAuthHolder(): OAuthHolder {
+  return {
+    getControlPlaneToken: () => "cp-token",
+    getDataPlaneToken: () => "dp-token",
+  } as unknown as OAuthHolder;
+}
+
+describe("oauth-client-manager.ts", () => {
+  describe("OAuthClientManager", () => {
+    describe("constructor", () => {
+      it("should derive the cloud REST URL from the Auth0 env (devel)", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "devel");
+        expect(cm["confluentCloudBaseUrl"]).toBe(
+          "https://api.devel.cpdev.cloud",
+        );
+      });
+
+      it("should derive the cloud REST URL from the Auth0 env (stag)", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "stag");
+        expect(cm["confluentCloudBaseUrl"]).toBe(
+          "https://api.stag.cpdev.cloud",
+        );
+      });
+
+      it("should derive the cloud REST URL from the Auth0 env (prod)", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "prod");
+        expect(cm["confluentCloudBaseUrl"]).toBe("https://api.confluent.cloud");
+      });
+
+      it("should reuse the cloud URL for the Tableflow base URL", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "stag");
+        expect(cm["confluentCloudTableflowBaseUrl"]).toBe(
+          cm["confluentCloudBaseUrl"],
+        );
+      });
+
+      it("should leave data-plane endpoints undefined (filled in by a later layer)", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "stag");
+        expect(cm["confluentCloudFlinkBaseUrl"]).toBeUndefined();
+        expect(cm["confluentCloudSchemaRegistryBaseUrl"]).toBeUndefined();
+        expect(cm["confluentCloudKafkaRestBaseUrl"]).toBeUndefined();
+        expect(cm["confluentCloudTelemetryBaseUrl"]).toBeUndefined();
+      });
+
+      it("should eagerly materialize the cloud REST client at startup", () => {
+        const eagerSpy = vi.spyOn(
+          BaseClientManager.prototype,
+          "getConfluentCloudRestClient",
+        );
+        new OAuthClientManager(fakeOAuthHolder(), "stag");
+        expect(eagerSpy).toHaveBeenCalledOnce();
+      });
+    });
+
+    describe("getKafkaClient()", () => {
+      it("should throw the OAuth-specific error", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "devel");
+        expect(() => cm.getKafkaClient()).toThrow(
+          /Native Kafka client is not available under OAuth/,
+        );
+      });
+    });
+
+    describe("getAdminClient()", () => {
+      it("should reject with the OAuth-specific error", async () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "devel");
+        await expect(cm.getAdminClient()).rejects.toThrow(
+          /Native Kafka client is not available under OAuth/,
+        );
+      });
+    });
+
+    describe("getProducer()", () => {
+      it("should reject with the OAuth-specific error", async () => {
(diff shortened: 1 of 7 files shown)


src/confluent/oauth-client-manager.test.ts
```diff
@@ -0,0 +1,119 @@
...
+          /Schema Registry SDK client is not available under OAuth/,
+        );
+      });
+
+      it("should not surface the BaseClientManager api_key-only error message", () => {
+        const cm = new OAuthClientManager(fakeOAuthHolder(), "devel");
+        expect(() => cm.getSchemaRegistryClient()).not.toThrow(
+          /SCHEMA_REGISTRY_API_KEY/,
+        );
+      });
```
shouples: This one could use a comment - are we just testing that `SCHEMA_REGISTRY_API_KEY` isn't included in the error thrown in the test above? (If so, does this need its own test?)

src/server-runtime.ts
```diff
@@ -58,10 +59,14 @@ export class ServerRuntime {
 
     // Construct a ClientManager for each connection in the config.
     // (although currently there will only be one, see `enforceSingleConnectionOnly()`)
+    // When OAuth is in play, every connection's manager is bearer-auth-backed by
+    // the shared holder; otherwise, fall back to the per-connection direct factory.
```
flippingbits, line 63: For clarification: Does this imply that we support only a single OAuth session at a time?

src/confluent/oauth-client-manager.ts
```diff
@@ -0,0 +1,98 @@
+/**
+ * @fileoverview OAuth (bearer-auth) client manager. Wires Confluent Cloud REST
+ * surfaces to bearer tokens supplied by an {@link OAuthHolder}; the cloud REST
+ * URL is auto-derived from the Auth0 environment via {@link getCloudRestUrlForEnv}.
+ * No native Kafka client — those tools are gated off by the `hasKafka` predicate
+ * (the synthesized OAuth connection has no `kafka` block); the Kafka methods
+ * below throw defensively if they're ever reached.
```
flippingbits, line 7: Do we plan to support these clients in OAuth connections at a later point, if possible?

flippingbits (approved):
Nice work!

> Telemetry endpoint stays unset under OAuth, intentionally: predicate gating prevents telemetry tools from being enabled under OAuth (no telemetry block in the synth connection). Once a future layer lets users supply a telemetry endpoint via YAML for OAuth connections, it'll flow through naturally.

Could we infer the Telemetry endpoint from the CCloud REST URL?

src/server-runtime.ts
bcscc: Yes, `enforceSingleConnectionOnly()` from https://github.com/confluentinc/mcp-confluent/pull/251 currently caps us at one connection for now while the migration is in progress, and even when multi-connection lands the holder is constructed once and shared across all per-connection managers, so a single OAuth session is intentional for now.

flippingbits: Makes sense! Should we then throw an error when users have configured multiple OAuth connections in the multi-connection config yaml? cc @jlrobins

src/confluent/oauth-client-manager.ts
bcscc: Yes, the SR and Kafka client tools require adding`librdkafka` to supports the OAUTHBEARER SASL mechanism so these throws are placeholders for "use the REST surface today, SDK + native Kafka coming via separate efforts."

bcscc:
> Nice work!
> 
> > Telemetry endpoint stays unset under OAuth, intentionally: predicate gating prevents telemetry tools from being enabled under OAuth (no telemetry block in the synth connection). Once a future layer lets users supply a telemetry endpoint via YAML for OAuth connections, it'll flow through naturally.
> 
> Could we infer the Telemetry endpoint from the CCloud REST URL?



> Nice work!
> 
> > Telemetry endpoint stays unset under OAuth, intentionally: predicate gating prevents telemetry tools from being enabled under OAuth (no telemetry block in the synth connection). Once a future layer lets users supply a telemetry endpoint via YAML for OAuth connections, it'll flow through naturally.
> 
> Could we infer the Telemetry endpoint from the CCloud REST URL?

Will add that support in the follow up. Ticketed [here](https://github.com/confluentinc/mcp-confluent/issues/311)

Issue #231, opened 2026-04-22
Migrate flink handlers off global env reads


jlrobins:
Replace FLINK_ENV_NAME, FLINK_DATABASE_NAME, FLINK_COMPUTE_POOL_ID, FLINK_ENV_ID env reads in flink handlers with `runtime.config.getSoleConnection().flink` fields. 

For example ...

`list-flink-statements-handler.ts` reads `env.FLINK_COMPUTE_POOL_ID` as a Zod `.default()`
callback on the `computePoolId` argument, and reads `env.FLINK_ENV_ID` directly inside
`handle()` as a fallback when `environmentId` is not supplied.

Both map to fields on `conn.flink`: `compute_pool_id` and `environment_id`.

## What changes

Remove the Zod `.default()` callback for `computePoolId`. Read `conn.flink.compute_pool_id` and
`conn.flink.environment_id` inside `handle()` via `runtime.config.getSoleConnection()` as
fallbacks when the corresponding tool arguments are absent. Remove `import env from "@src/env.js"`.

## Tests

Add `handle()` tests covering:

- Explicit tool argument values take precedence over connection config
- Absent arguments fall back to connection config values
- Stub `ClientManager` with `createMockInstance(DefaultClientManager)`

PR #302, opened 2026-05-01
Extract reusable handle() test infrastructure; add two POC focused handler suites

jlrobins:
## Summary of Changes

**TL;DR:** Promotes the two-proxy handler stub machinery out of the smoke test and into `tests/stubs/`, adding typed harness utilities that reduce a per-handler `handle()` test suite to a single `HandleCase[]` array — no repeated assertion logic, no per-case boilerplate. Closes the follow-on noted in PR #295.

Two proof-of-concept `handle()` suites ship with this PR. `list-flink-statements-handler` reaches 88.9% statement and 83.3% branch coverage in 4 cases; `get-flink-exceptions-handler` reaches 85.7% statement coverage in 5 cases. Both suites are pure data tables — label, args, optional `responseData` shape, expected outcome — with zero structural variation between cases. The only lines left uncovered in each are the `if (error)` API-error branches, which the proxy cannot reach because it always resolves `error` as `undefined`; those branches will become testable once the sequential-response extension described in the Future section is implemented.

## Intent

The `tool-registry.test.ts` smoke suite introduced a universal two-proxy `ClientManager` stub and an assertion helper (`assertHandleCase`) that proved useful enough to live somewhere reusable. Without extraction, every focused handler test suite would re-implement the proxy pair, the thrown-vs-resolved discriminator, and the `it.each` wiring independently. Issue #231 and its siblings will touch ~15 Flink handlers; the per-handler cost of writing `handle()` coverage should be "declare a data table," not "copy infrastructure."

## Changes by file category

**`tests/stubs/handler.ts`** (new) — the extracted core. Contains the two-proxy pair (`callableProxy` + `responseProxy`) and all harness types and utilities:

- `stubClientGetters(responseData?)` — wires every `DefaultClientManager` getter mock to the proxy pair; returns `{ clientManager, clientGetters }` for injection into `runtimeWith()` and `assertHandleCase()` respectively
- `assertHandleCase(options)` — runs `handler.handle(runtime, args)` and asserts against a `HandleOutcome` discriminated union; the `"TODO"` sentinel drives a discovery workflow that reports the correct entry to paste in
- `HandleOutcome = Resolves | Throws | "TODO"` — the discriminated union specifying what a case expects
- `HandleCase` — the `it.each` row shape: `{ label, args, outcome, responseData? }`
- `ClientGetters`, `Resolves`, `Throws`, `classifyThrown` — supporting types and utilities

**`tests/stubs/index.ts`** — re-exports all new symbols alongside the existing `admin`, `mock-instance`, `node-deps`, and `stub-handler` exports.

**`src/confluent/tools/tool-registry.test.ts`** — the smoke suite's per-tool assertion loop collapses from ~100 lines to ~8 by consuming the shared utilities. `allServicesRuntime`, `ZERO_ARG_OUTCOMES`, and `TOOL_NAME_TO_KEY` remain local — they are smoke-suite concerns, not general infrastructure.

**`src/confluent/tools/handlers/flink/list-flink-statements-handler.test.ts`** (new) — four-case `handle()` suite covering the missing-org-ID guard, successful resolution with explicit args, client-side `statusPhase` filtering with a mixed-phase result set, and the zero-match filter case. The `responseData` field is shaped to push the handler past the API call and into its filtering logic rather than stopping at the guard layer.

**`src/confluent/tools/handlers/flink/get-flink-exceptions-handler.test.ts`** (new) — five-case suite covering Zod validation failure (absent `statementName`), the two sequential `getEnsuredParam` guard failures (org then env), the empty-exceptions success path, and the populated-exceptions success path.

Both new suites carry a `beforeAll(() => { initEnv(); })` with a comment noting it can be removed once issue #231 eliminates the Zod `.default(env.FLINK_*)` and `getEnsuredParam` env reads from the handlers.

## Future

`stubClientGetters` currently returns the same `responseData` for every API call, which is sufficient for handlers that make a single REST call or that are only being tested at the guard layer. Multi-call handlers (`check-health`, `detect-issues`, `read-flink-statement`, `query-profiler`) will eventually need per-call response sequencing. The planned extension: accept `responseData` as an array, making the `apply` trap index-aware and constructing a fresh `responseProxy` per invocation — backward-compatible, localized entirely inside `stubClientGetters`. A comment in the source flags the spot.

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes

### DIFF
--- a/src/confluent/tools/handlers/flink/get-flink-exceptions-handler.test.ts
+++ b/src/confluent/tools/handlers/flink/get-flink-exceptions-handler.test.ts
@@ -1,0 +1,101 @@
+import { GetFlinkExceptionsHandler } from "@src/confluent/tools/handlers/flink/get-flink-exceptions-handler.js";
+import { initEnv } from "@src/env.js";
+import {
+  DEFAULT_CONNECTION_ID,
+  runtimeWith,
+} from "@tests/factories/runtime.js";
+import {
+  assertHandleCase,
+  stubClientGetters,
+  type HandleCase,
+} from "@tests/stubs/index.js";
+import { beforeAll, describe, it } from "vitest";
+
+const FLINK_CONN = {
+  flink: {
+    endpoint: "https://flink.example.com",
+    auth: { type: "api_key" as const, key: "k", secret: "s" },
+    environment_id: "env-from-config",
+    organization_id: "org-from-config",
+    compute_pool_id: "lfcp-from-config",
+  },
+};
+
+const EXPLICIT_IDS = {
+  organizationId: "org-from-args",
+  environmentId: "env-from-args",
+};
+
+const STATEMENT_NAME = "my-statement";
+
+describe("get-flink-exceptions-handler.ts", () => {
+  describe("GetFlinkExceptionsHandler", () => {
+    const handler = new GetFlinkExceptionsHandler();
+
+    // Required while the handler reads env vars via getEnsuredParam.
+    // Remove after issue #231 migrates those reads to conn.flink config fields.
+    beforeAll(() => {
+      initEnv();
+    });
+
+    describe("handle()", () => {
+      const cases: HandleCase[] = [
+        {
+          label: "throws when statementName is absent",
+          args: {},
+          outcome: { throws: "ZodError" },
+        },
+        {
+          label: "throws when organizationId is absent and not in env",
+          args: { statementName: STATEMENT_NAME },
+          outcome: { throws: "Organization ID is required" },
+        },
+        {
+          label: "throws when environmentId is absent and not in env",
+          args: {
+            statementName: STATEMENT_NAME,
+            organizationId: "org-from-args",
+          },
+          outcome: { throws: "Environment ID is required" },
+        },
+        {
+          label: "resolves with no-exceptions message when data is empty",
+          args: { statementName: STATEMENT_NAME, ...EXPLICIT_IDS },
+          responseData: { data: [] },
+          outcome: {
+            resolves: `No exceptions found for statement '${STATEMENT_NAME}'.`,
+          },
+        },
+        {
+          label: "resolves with exception list when exceptions are present",
+          args: { statementName: STATEMENT_NAME, ...EXPLICIT_IDS },
+          responseData: {
+            data: [{ message: "OOM error" }, { message: "Timeout" }],
+          },
+          outcome: {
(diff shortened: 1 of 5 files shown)


tests/stubs/handler.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  if (outcome === "TODO") {
+    const discovered =
+      thrown === undefined
+        ? `{ resolves: "${result!.content
+            .map((c) => ("text" in c ? c.text : ""))
+            .join("")
+            .slice(0, 60)}" }`
+        : `{ throws: "${classifyThrown(name, thrown)}" }`;
+    throw new Error(`${name}: outcome is TODO — replace with: ${discovered}`);
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/confluent/tools/handlers/flink/get-flink-exceptions-handler.test.ts
```diff
@@ -0,0 +1,101 @@
...
+            data: [{ message: "OOM error" }, { message: "Timeout" }],
+          },
+          outcome: {
+            resolves: `Flink Statement Exceptions for '${STATEMENT_NAME}'`,
+          },
+        },
+      ];
+
+      it.each(cases)(
+        "should $label",
```
shouples, line 82: Minor: it might be easier just putting the entire test name in the `label` since it currently breaks these out into:
- `should resolves with ...`
- `should throws when ...`

src/confluent/tools/handlers/flink/list-flink-statements-handler.test.ts
```diff
@@ -0,0 +1,99 @@
...
+          },
+          responseData: {
+            data: [{ status: { phase: "FAILED" } }],
+          },
+          outcome: { resolves: "Found 0 statement(s) with status 'RUNNING'" },
+        },
+      ];
+
+      it.each(cases)(
+        "should $label",
```
shouples, line 80: Same here ^

shouples (approved):
Nice cleanup, just one minor suggestion for the test names

src/confluent/tools/handlers/flink/list-flink-statements-handler.test.ts
jlrobins: Will revise in the next PR in the stack, thanks!

PR #345, opened 2026-05-05
feat: Confluent Platform support (minimal scope)

rohit2b:
## Summary

Make the existing Kafka admin and Schema Registry tools work cleanly against a self-managed Confluent Platform cluster, without introducing new modes, flags, or auth paths. CP support is expressed as a different set of values for connection blocks the existing predicate-based gating already understands — no new architecture.

This PR also fixes a real under-gating bug surfaced while building it: the `hasCCloudCatalogSupport` predicate at `src/confluent/tools/connection-predicates.ts:48-50` returned true whenever `schema_registry.auth.type === "api_key"` — a shape that matches both Confluent Cloud SR and a self-managed CP SR fronted by HTTP Basic Auth. The latter doesn't serve `/catalog/v1/`, so the seven Stream Catalog tools (`catalog/*` + `search/search-topic-by-tag` + `search/search-topics-by-name`) over-enabled on CP and 404'd at runtime. The predicate now also requires a `confluent_cloud` block, which CCloud users have and CP users don't.

## Changes

| Commit | Purpose |
|---|---|
| `chore(git)` | Ignore `.worktrees/` for local feature branches |
| `docs` | Sample CP YAML config in `sample_configs/` |
| `test(cp)` | Local CP docker-compose stack (KRaft + SR) for integration tests |
| `docs(env)` | Clarify deployment-neutral env-var descriptions in `src/env-schema.ts` |
| **`fix(tools)`** | **Predicate fix: `hasCCloudCatalogSupport` requires `confluent_cloud` block** |
| `test(tools)` | Sweep test guarding the categorical CP-gating invariant in one place |
| `test(cp)` | Seven per-handler `*.cp.integration.test.ts` files, `@cp` tag, harness modules, fixture YAML, `.env.integration.example` keys |
| `docs(readme)` | "Using with Confluent Platform" section: env vars, SCRAM escape hatch, NODE_EXTRA_CA_CERTS, log-based verification, smoke-test stack |

## Behavior change to flag in release notes

Existing CCloud users who set `SCHEMA_REGISTRY_*` env vars / blocks but **not** `CONFLUENT_CLOUD_API_KEY/SECRET` (or no `confluent_cloud` block in YAML) will see seven Stream Catalog tools auto-disable at startup with a "no connections satisfy its requirements" log line, where they previously registered and 404'd at first call. Strict UX improvement, but worth surfacing.

## Test plan

- [x] `npm run test:unit` — 946 passing (62 files)
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] CP gating sweep test (`src/confluent/tools/cp-gating.test.ts`) — proves the categorical invariant for every registered handler
- [x] Predicate test — added a new case for "schema_registry has api_key auth but confluent_cloud is absent" returning false (Confluent Platform with auth-protected SR)
- [ ] **Manual:** start the docker-compose stack and run the @cp integration suite
  ```bash
  docker compose -f docker-compose.cp-test.yml up -d
  CP_KAFKA_USERNAME=mcp CP_KAFKA_PASSWORD=mcp-secret \
    npm run test:integration -- --tags-filter=@cp
  docker compose -f docker-compose.cp-test.yml down -v
  ```
  (Integration tests added in this PR are skip-by-default and were not run against Docker locally.)
- [ ] **Manual:** start the server briefly with a CP config and confirm the per-tool startup logs name the expected disabled tools.

## Out of scope (filed as follow-ups)

- Rerouting `alter-topic-config`, `get-topic-config`, `search-topics-by-name` to KafkaJS admin (currently disabled on CP because they require `kafka.rest_endpoint`)
- First-class SASL-SCRAM support (works today via the `extra_properties` escape hatch, documented as unofficial)
- CP-specific tooling (MDS/RBAC, Cluster Linking, ksqlDB, CP Flink, Control Center)


sample_configs/confluent-platform.yaml
```diff
@@ -0,0 +1,36 @@
...
+# TLS trust: if your CP brokers or Schema Registry use an internal CA not in Node's
+# default trust store, set NODE_EXTRA_CA_CERTS=/path/to/ca.pem in your environment.
+
+connections:
+  cp:
+    type: "direct"
+    kafka:
+      bootstrap_servers: "cp-broker.internal:9092"
+      auth:
+        type: "api_key"
```
jlrobins, line 23: Would it be cleaner overall if we expanded auth: type: beyond just "api_key" here?

jlrobins (commented):
One initial comment about your opinion as to the modeling of these alternate auth types, and am curious if you found the testing and tooling guidance markers recently added in the codebase as helpful.

README.md
```diff
@@ -78,6 +79,70 @@ SCHEMA_REGISTRY_ENDPOINT="http://localhost:8081"
 | **Schema Registry** | `list-schemas`, `delete-schema`                                                        | List, inspect, and delete data schemas            |
 | **Documentation**   | `search-product-docs`                                                                  | Search Confluent product documentation by keyword |
 
+## Using with Confluent Platform
+
+`mcp-confluent` runs against a self-managed Confluent Platform (CP) cluster the same way it runs against any local Kafka + Schema Registry: by configuring a connection that has only the service blocks the deployment actually exposes. There is no separate "CP mode" — tool enablement is decided per-connection by the predicates in [`src/confluent/tools/connection-predicates.ts`](src/confluent/tools/connection-predicates.ts), and a connection without a `confluent_cloud` block automatically gates out every Cloud-only tool (Flink, Tableflow, Billing, Metrics, Stream Catalog tags, Managed Connect, Organizations / Environments / Clusters).
```
jlrobins: End-users probably do not need to know the internal code implementation details.

jlrobins: Also, the handle() test infrastructure has changed. Check out https://github.com/confluentinc/mcp-confluent/pull/360#issuecomment-4389344028 for notes on how to easily upgrade to the latest and greatest, thanks!

bcscc:
Hey @rohit2b, would love to get this in, just need some updates on the testing infrastructure as mentioned [above](https://github.com/confluentinc/mcp-confluent/pull/345#discussion_r3196491639). We've updated the [.claude/rules](https://github.com/confluentinc/mcp-confluent/tree/main/.claude/rules) files so the updates should be straight-forward.  However happy to directly make the fix on this PR if needed!

bcscc (approved):

PR #363, opened 2026-05-06
Cleanup/305 delete get ensured param and test cleanup

### LINKED ISSUE #305: Clean up any remaining uses of getEnsuredParam() and delete it
getEnsuredParam() indirectly consults the global envs. Find any remaining callers, then refactor was done for the Flink handlers in #231, then delete the function.

Ensure there is handle() test coverage over all the affected codepaths:
  * when the parameter is provided in the tool call,
  * when it isn't but is found in the configuration,
  * when it isn't and is NOT defaultable from the configuration.

Now is also a good time to take stock on all of the changes done to get to this point and to identify any replicated patterns in either the test suite or codebase proper to then centralize.

jlrobins:
## Summary

- Deletes `getEnsuredParam()` and removes the file `src/confluent/helpers.ts` (which contained nothing else). The function was the last vestige of global-env consultation in the handler resolution path, but had no callers left after the migration sequence — this PR closes the loop by removing the dead code.
- Rewrites a stale doc-comment in `src/config/models.ts` that still described `cluster_id` / `env_id` resolution as "via `getEnsuredParam`, which checks (in order) the tool argument supplied by the LLM, then the env var, then throws". Both the function name and the env-var step are now obsolete; the new comment describes the per-handler resolver pattern accurately without naming a specific helper.
- Verified `handle()` test coverage for the three cases the issue calls out — arg supplied, arg absent + config fallback, arg absent + no config → throws — on every handler that previously called `getEnsuredParam`. Tableflow (8 handlers from #353), connect (4 from #350), and kafka topic-config (2) all already had the throw case. The seven Flink handlers do not — and intentionally won't. Their connection-block Zod schema requires `organization_id`, `environment_id`, and `compute_pool_id` (each `.trim().min(1)` or `.startsWith(...)`), and the `hasFlink` predicate gates tool enablement on the block's presence, so the throw inside `resolveParam` is provably unreachable in production for Flink. Adding tests would simulate a Zod-bypassed runtime rather than real handler behaviour.

## While here

Three rounds of type-alias deduplication across handler tests. The trigger was a late review note on #353 about a copy-pasted `TableflowHandleCase`; surveying the rest of the suite surfaced two more instances of the same anti-pattern.

- `TableflowHandleCase` was declared identically in eight tableflow tests. Hoisted to `tests/factories/runtime.ts` next to its sibling `HandleCaseWithConn` and the `TABLEFLOW_CONN` fixture; tests now import the shared type.
- `ConnectHandleCase` had the same shape duplicated across four connect tests. Same remedy. The verb-specific JSDoc on each local copy (GET / POST / DELETE) collapses into a single verb-agnostic JSDoc on the shared type.
- Six Flink GET tests each invented a fresh name for the same `{ flinkGetData?: unknown }` shape: `ExceptionsCase`, `QueryProfilerCase`, `ReadStatementCase`, `IssuesCase`, `CheckHealthCase`, and `ListStatementsCase`. Replaced with a single shared `FlinkGetCase`. (`ListStatementsCase` had `flinkGetData: unknown` required rather than optional; the optional variant in the shared type works fine since the test always supplies it.)

Single-use case types deliberately kept as-is — one declaration each, nothing to deduplicate: `DeleteStatementCase` (Flink DELETE; status-code-flavoured rather than body-flavoured), `EnvIdCase` (clusters), `FilterCase` (metrics).

Revised unit-tests.md accordingly with hint over preferring to centralize common HandleCase definitions.

## Relationships

- Closes #305

## Pull request checklist

#### Tests

- [ ] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes


shouples (approved):
Oh now this is some nice cleanup 👏

Issue #358, opened 2026-05-05
Add `describe-tool-availability` tool to surface per-connection enablement verdicts


jlrobins:
### Problem

When a user misconfigures their connection — missing a required block, wrong field name,
incomplete credentials — tools silently disappear from `tools/list` with no explanation.
The server knows exactly why each tool is absent, but that knowledge is not surfaced
anywhere the user or an AI assistant can see it.

### Proposed design

Add a new MCP tool, `describe-tool-availability`, that returns the full enablement verdict
for every registered tool across every configured connection. The tool requires no arguments
and is always enabled (it has no connection-block requirement of its own).

Response shape (one entry per tool):

```typescript
{
  tool: string;           // ToolName enum value
  enabled_for: string[];  // connection IDs where the tool is active
  disabled_for: {
    connection_id: string;
    reason: string;
  }[];
}
```

An AI assistant receiving this output can tell the user precisely what to add to their
config to unlock a given tool — rather than leaving them to guess why it isn't appearing.

As the server grows to support multiple connections, this tool becomes the canonical way
to reason about the difference sets between connections: which tools a given connection
enables that others don't, and exactly what config each connection is missing to reach
parity.

### Example output (excerpt)

```
list-tableflow-topics
  enabled_for: []
  disabled_for:
    - default: "no tableflow block in connection config"

create-flink-statement
  enabled_for: ["default"]
  disabled_for: []
```

Issue #357, opened 2026-05-05
Handler layer: replace `enabledConnectionIds()` overrides with `readonly predicate` property


jlrobins:
### What

Refactor `BaseToolHandler` so that the coordination between a handler's predicate and
the enabled-connection machinery lives in one place. Every domain base class and
individual handler currently implements `enabledConnectionIds()` as a one-liner that
closes over a predicate; after this change they declare only a `readonly predicate`
property and the base class derives both `enabledConnectionIds()` and
`disabledConnectionReasons()` from it.

### Scope

~32 files. The per-handler changes are a uniform find-and-replace; the risk is breadth,
not complexity.

### Checklist

- [ ] Add `abstract readonly predicate: (conn: ConnectionConfig) => PredicateResult` to `BaseToolHandler`
- [ ] Make `enabledConnectionIds()` concrete on `BaseToolHandler`, implemented via `connectionIdsWhere(runtime.config.connections, this.predicate)`
- [ ] Add `disabledConnectionReasons(runtime): Map<string, PredicateResult>` to `BaseToolHandler`, implemented via `connectionReasonsWhere(runtime.config.connections, this.predicate)`
- [ ] Replace `enabledConnectionIds()` override with `readonly predicate = hasXxx` in all 3 domain base classes (`FlinkToolHandler`, `TableflowOnlyToolHandler`, `TableflowWithKafkaToolHandler`, `ConnectToolHandler`)
- [ ] Same replacement in all ~27 individual handlers that currently override `enabledConnectionIds()` directly
- [ ] Base-case handlers with no domain requirement get `readonly predicate = (_: ConnectionConfig): PredicateResult => ({ enabled: true })`
- [ ] Add startup logging in `ServerRuntime.fromConfig()` using `disabledConnectionReasons()`, grouped by `{ connectionId, reason }` with all affected tool names listed together — one log entry per distinct reason, not one per tool

Issue #355, opened 2026-05-05
Replace boolean connection predicates with verdict objects that carry a reason


jlrobins:
### Problem

Connection predicates in `connection-predicates.ts` return plain `boolean`. When a predicate
returns `false`, `connectionIdsWhere()` silently omits the connection from the enabled set and
the tool disappears from `tools/list` with no explanation.

This is already better than the old env-var hodgepodge (where tools appeared but threw at
runtime with cryptic missing-env errors), but it creates a discoverability gap: a user who
configures a `tableflow` block but forgets `kafka.env_id` will find that all eight
env-resolving Tableflow tools simply don't appear. Nothing in the server output explains why.

### Proposed design

Replace the `boolean` return type with a verdict object:

```typescript
export type PredicateResult =
  | { enabled: true }
  | { enabled: false; reason: string };
```

Each predicate returns a `PredicateResult` instead of `boolean`, recording exactly which
condition failed. `connectionIdsWhere()` continues to return `string[]`; a companion
`connectionReasonsWhere()` returns `Map<connectionId, PredicateResult>` for diagnostic use.

Each handler declares a `readonly predicate` property; `BaseToolHandler` derives both
`enabledConnectionIds()` and `disabledConnectionReasons()` from it, eliminating the
per-handler boilerplate that exists today.

### What this unlocks

- **Startup logging.** `ServerRuntime.fromConfig()` can log one entry per distinct failure
  reason, listing all affected tool names together — making misconfiguration immediately
  visible without flooding the log.
- **Diagnostic tool.** A future `describe-tool-availability` tool or resource could return
  the full verdict map so an AI assistant can tell the user exactly what to add to their
  config to unlock a tool.
- **Self-documenting predicates.** The reason strings make each predicate's failure modes
  legible without consulting external docs.

### Timing

Both child issues land after #234 (env retirement), the final issue in the #230 migration.

### Relationships

- Part of the broader env-to-config migration tracked in #230
- Depends on #234 (full env retirement) for the natural retrofit window
- Informed by the predicate invariant established during #348 (Tableflow) and #350 (connectors)
- Tracked in child issues: #356 (predicate layer) and #357 (handler layer)

This work supersedes earlier authored issue #288, but is exactly the same spirit.

PR #369, opened 2026-05-06
Expand predicate results to describe reason why failed.

### LINKED ISSUE #356: Predicate layer: introduce `PredicateResult` and `connectionReasonsWhere()`
### What

Replace the `boolean` return type of all connection predicates in
`connection-predicates.ts` with a `PredicateResult` verdict object that records
exactly which condition failed. Add a companion `connectionReasonsWhere()` function
that returns the full verdict map for a connection set.

### Scope

~3 files. No handler files are touched — `connectionIdsWhere()` absorbs the type
change internally and still returns `string[]`, so nothing upstream of it sees
`PredicateResult`.

### Checklist

- [ ] Add `PredicateResult` discriminated union type to `connection-predicates.ts`
- [ ] Update all predicates to return `PredicateResult` with a specific `reason` string on each false branch
- [ ] Update `connectionIdsWhere()` to filter on `.enabled`
- [ ] Add `connectionReasonsWhere(connections, predicate): Map<string, PredicateResult>`
- [ ] Update `connection-predicates.test.ts` to assert `{ enabled: true }` / `{ enabled: false, reason: "..." }` on every case — pin the exact reason string, not just `enabled: false`

jlrobins:
## Predicate verdicts

- Replaces the `boolean` return type of every connection predicate in `connection-predicates.ts` with a `PredicateResult` discriminated union: `{ enabled: true } | { enabled: false; reason: ToolDisabledReason }`. Predicates now report exactly which condition failed; the previous `false` was a black hole that left misconfigured users staring at silently-missing tools.
- Centralises every failure reason in a `ToolDisabledReason` enum whose value _is_ the human-readable phrasing. Source consumers reference the symbol (compile-time stable); #357's startup-log grouping and #358's `describe-tool-availability` diagnostic tool will surface the values directly. Embedding the message in the enum beats a parallel lookup record — one source of truth, no drift seam, and the docstring on the enum is the single roof under which every reason lives. The phrasings are deliberately first-pass; #357/#358 will revise as the rendering layer materialises.
- Compound predicates (`hasKafkaRestWithAuth`, `hasCCloudCatalogSupport`) return layered/specific reasons (`MissingKafkaRestEndpoint` then `MissingKafkaAuth`, etc.) rather than a single combined reason. The enum makes layered-and-specific as cheap to express as combined-and-vague, and #358's "tell the user what to add" story benefits from the granularity.
- Adds `connectionReasonsWhere(connections, predicate): Map<string, PredicateResult>` alongside the existing `connectionIdsWhere`. The map preserves insertion order and includes both enabled and disabled connections, so #357 can iterate it once to build startup-log groupings and #358 can return it as the canonical diagnostic shape. `connectionIdsWhere` keeps its `string[]` signature; the type change is absorbed internally (`predicate(conn).enabled`).
- Tests pin the exact `ToolDisabledReason` symbol on every disabled case — phrasing edits to enum values won't ripple through ~50 assertions. Verdict literals collapse to `toEqual(ENABLED)` and `toEqual(disabledFor(ToolDisabledReason.X))` via an exported `ENABLED` constant and a test-suite-local `disabledFor()` helper, so each assertion reads as a one-liner that names its intent. Added a brand-new `hasDirectConfluentCloud()` describe block (that predicate had zero direct test coverage before this pass) and four new `connectionReasonsWhere` cases (empty map, mixed verdicts with order pinned, all-enabled, all-disabled).

## Predicate AND-collapse and the `allOf` combinator

- The type change introduced a sharp edge: `predA(conn) && predB(conn)` AND'd two booleans before, but both operands are now truthy `PredicateResult` objects. The `&&` silently collapses to the second operand and `connectionIdsWhere` only inspects the second predicate's `.enabled` — dropping the first check entirely. TypeScript is content because the lambda still returns a `PredicateResult`. **Two call sites used this pattern.** Claude caught `QueryProfilerHandler.enabledConnectionIds` (`hasFlink && hasTelemetry`) during the original refactor pass; Copilot's review caught a second instance Claude missed and had explicitly asserted didn't exist: `CreateConnectorHandler.enabledConnectionIds` (`hasDirectConfluentCloud && hasKafkaAuth`), same shape, same blast radius.
- Introduced `allOf(...predicates)` as an exported combinator. Returns `ENABLED` only when every predicate passes; otherwise the first failing verdict, so the specific reason still propagates to `connectionReasonsWhere`. Both call sites now use `allOf`; the helper's docstring documents that raw `&&` between predicates is wrong, so future readers won't reintroduce the pattern.
- Pinned both bugs with regression tests on the respective handlers — each exercises the case that fails _only_ under the buggy collapse (a connection satisfying the second predicate but not the first). For each test, verified it goes red against a deliberately-buggy version of the lambda before re-applying the `allOf` fix, so the test is proven to catch the exact regression. `allOf` itself is covered by five unit tests in `connection-predicates.test.ts` (all-pass, first-fails, later-fails, short-circuit no-call-after-failure, zero-predicate vacuous-true).

## Relationships

- Closes #356
- First child of #355 — the handler-layer refactor (#357) and `describe-tool-availability` MCP tool (#358) build on the verdict shape this PR introduces.

### DIFF
--- a/src/confluent/tools/connection-predicates.test.ts
+++ b/src/confluent/tools/connection-predicates.test.ts
@@ -1,9 +1,14 @@
 import type { DirectConnectionConfig } from "@src/config/index.js";
 import type { ConnectionConfig } from "@src/config/models.js";
 import {
+  ENABLED,
+  type PredicateResult,
+  ToolDisabledReason,
   connectionIdsWhere,
+  connectionReasonsWhere,
   hasCCloudCatalogSupport,
   hasConfluentCloud,
+  hasDirectConfluentCloud,
   hasFlink,
   hasKafka,
   hasKafkaAuth,
@@ -19,6 +24,10 @@ function conn(
   fields: Omit<DirectConnectionConfig, "type">,
 ): DirectConnectionConfig {
   return { type: "direct", ...fields };
+}
+
+function disabledFor(reason: ToolDisabledReason): PredicateResult {
+  return { enabled: false, reason };
 }
 
 const KAFKA_CONN = conn({ kafka: { bootstrap_servers: "broker:9092" } });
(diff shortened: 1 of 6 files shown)


src/confluent/tools/connection-predicates.ts
```diff
@@ -1,85 +1,249 @@
...
-// Block-presence predicates below check `conn.type === "direct"` first; OAuth
-// connections carry no service blocks, so they answer false for every block.
-// `hasConfluentCloud` is the single exception: it widens for OAuth because the
-// CCloud REST URL is reachable via the Auth0 environment without a block.
+export type ConnectionPredicate = (conn: ConnectionConfig) => PredicateResult;
 
-export function hasKafka(conn: ConnectionConfig): boolean {
-  return conn.type === "direct" && conn.kafka !== undefined;
+export const ENABLED: PredicateResult = { enabled: true };
+
```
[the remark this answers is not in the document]
jlrobins: Excellent suggestion, Copilot! Fixed!

src/confluent/tools/connection-predicates.test.ts
```diff
@@ -13,14 +19,18 @@ import {
...
 
 function conn(
   fields: Omit<DirectConnectionConfig, "type">,
 ): DirectConnectionConfig {
   return { type: "direct", ...fields };
 }
 
+function disabledFor(reason: ToolDisabledReason): PredicateResult {
+  return { enabled: false, reason };
+}
```
shouples, line 32: Any reason we can't just export and use `disabled()` directly in the tests?

src/confluent/tools/connection-predicates.ts
```diff
@@ -1,85 +1,272 @@
...
+  MissingKafkaBootstrap = "'kafka' block does not have 'bootstrap_servers' field",
+  MissingKafkaAuth = "'kafka' block does not have 'auth' field",
+  MissingKafkaRestEndpoint = "'kafka' block does not have 'rest_endpoint' field",
+  MissingSchemaRegistryBlock = "no 'schema_registry' block in connection config",
+  MissingSchemaRegistryApiKeyAuth = "'schema_registry' block does not have 'auth' field of type 'api_key'",
+  MissingConfluentCloudBlock = "no 'confluent_cloud' block in connection config",
+  MissingFlinkBlock = "no 'flink' block in connection config",
+  MissingTelemetryBlock = "no 'telemetry' block in connection config",
+  MissingTableflowBlock = "no 'tableflow' block in connection config",
+  OAuthNoServiceBlocks = "OAuth connections carry no service blocks",
```
shouples, line 270: Do we expect this wording to change as tools are migrated to support OAuth connections?

src/confluent/tools/connection-predicates.ts
```diff
@@ -1,85 +1,272 @@
...
+ * present and its `auth` field is `api_key`-typed. That combination is the
+ * reliable signal that the SR is CCloud-hosted and therefore exposes the
+ * `/catalog/v1/` endpoints. A vanilla CP SR has no auth, so it disables
+ * even when the `schema_registry` block itself is present.
+ */
+export function hasCCloudCatalogSupport(
+  conn: ConnectionConfig,
+): PredicateResult {
+  if (conn.type === "oauth")
+    return disabled(ToolDisabledReason.OAuthNoServiceBlocks);
```
shouples, line 192: Similarly, do we expect these predicates to not-early-exit once tool handlers migrate to support OAuth connections?

shouples (approved):
Looks good from my perspective, just one question and one potential area for deduplication for tests

src/confluent/tools/connection-predicates.test.ts
jlrobins: I thought about that, but then if disabled() itself was grossly wrong, then testing it ... with itself is bad. Duplicating its expected shape is better route IMO.

src/confluent/tools/connection-predicates.ts
jlrobins: Yes, for sure. These are initial strings to get us rolling. Product et all should have a hand in revising.

src/confluent/tools/connection-predicates.ts
jlrobins: Yes, but probably in @bcscc court to concurrently revise as he spans out.

PR #379, opened 2026-05-07
Add --init-config flag to bootstrap a starter config.yaml

### LINKED ISSUE #378: Add --init-config flag to bootstrap local config.yaml without cloning the repo
**Quality-of-life improvement for users who run the MCP server via `npx` / a global install and never check out this repo locally.**

### Problem

Today the documented bootstrap path for the YAML-based config is "clone or browse the repo, then `cp config.example.yaml config.yaml`" (`README.md:160-168`). That couples first-time setup to having a checkout of the source tree, which is at odds with the rest of our CLI surface — `--generate-key` already exists for users who only ever invoke the package via `npx @confluentinc/mcp-confluent`.

As we move users off the env-var path (#151) onto YAML, anyone running the published package with `npx` has to either copy-paste `config.example.yaml` from GitHub raw or check out the repo just to grab one file. Neither is friendly. This issue is squarely about that audience: people who install or `npx` the package and have no reason to clone the source tree.

### Proposed design

Add an early-exit CLI flag that writes a starter `config.yaml` into the user's current working directory:

```bash
npx @confluentinc/mcp-confluent --init-config
# wrote config.yaml (added to .gitignore; next: edit credentials and run with --config ./config.yaml)
```

Behavior:
- Default destination: `./config.yaml` in CWD.
- If the destination already exists, refuse and exit non-zero with an error message.
- Like `--generate-key`, runs before any env loading or YAML validation and `process.exit(0)` on success. Bootstrapping must not require the very file it's bootstrapping.

Wired into `src/cli.ts` as a new `CLIOptions.initConfig` field, handled in `src/index.ts` adjacent to the existing `cliOptions.generateKey` branch (`src/index.ts:100`), in a small helper alongside `outputApiKey()` (`src/index.ts:72`).

### Auto-gitignore the created file

The generated `config.yaml` will hold credentials (or `${VAR}` references that are easy to accidentally fill in-place). After writing the file, `--init-config` updates `.gitignore` in the same directory:

- If `<dir>/.gitignore` does not exist, create it containing the file's basename (e.g. `config.yaml`).
- If it exists, append the basename only when no existing line (after trimming) already matches it. Idempotent — re-running `--init-config` does not produce duplicate entries.
- The `.gitignore` lives next to the generated file.
- Run silently and report in the success message what was added (`wrote ./config.yaml (added to .gitignore)` vs. `wrote ./config.yaml (.gitignore already excludes it)`).

This complements — does not replace — the existing repo-root `.gitignore` rule that excludes all root-level `*.yaml`/`*.yml` files (`README.md:162`); that rule only protects users who happen to be bootstrapping inside the source checkout, which is exactly the audience this feature is *not* primarily aimed at.

### Where the file content comes from

The npm tarball today only ships `dist/` (`package.json#files: ["dist"]`), so `config.example.yaml` from the repo root is **not** present in the installed package. The fix is to add `config.example.yaml` to `package.json#files` and have the CLI read it relative to its own install location (`new URL("../config.example.yaml", import.meta.url)`). One source of truth, zero build wiring. The build pipeline is `tsc + tsc-alias` with no asset-copy step today, and `package.json#files` is the lowest-friction way to keep the example file authoritative for both repo users and npm consumers.

### Out of scope

- Interactive prompting for connection details. The example is annotated and the `${VAR}` interpolation flow covers the credentials path. A future `--init-config --interactive` is a possible follow-up.
- Generating the example from the Zod schema in `src/config/models.ts`. Would eliminate drift between schema and example, but is materially more work and not needed to close the bootstrap-without-repo gap.
- Detecting whether the destination directory is inside a git working tree before writing `.gitignore`. Creating an unused `.gitignore` outside a git repo is harmless and cheaper than walking up looking for `.git`; users who don't want it can pass `--no-gitignore`.
- Telemetry. The flag exits before `TelemetryService.initialize()`; no wiring needed.

### Context

- README workflow this replaces: `README.md:160-168`.
- Sibling early-exit flag for reference: `--generate-key` (`src/cli.ts:210`, handler at `src/index.ts:100`, helper at `src/index.ts:72`).
- Related: #151 (env-var → YAML migration).

flippingbits:
## Summary

- Adds `npx @confluentinc/mcp-confluent --init-config`: writes a starter `config.yaml` (a copy of the bundled `config.example.yaml`) into the user's CWD and adds it to a sibling `.gitignore` so credentials filled in later don't slip into git.
- Refuses to overwrite an existing `config.yaml` so a stray rerun won't clobber edits; gitignore append is idempotent.
- Ships `config.example.yaml` in the npm tarball (`package.json#files`) so users running via `npx` no longer need a checkout to bootstrap.

Fixes #378.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -91,21 +91,27 @@ export function outputApiKey(): void {
  * is the user's project but the example file lives next to the install.
  *
  * Refuses to overwrite an existing destination so an accidental rerun
- * cannot wipe credentials the user already filled in.
+ * cannot wipe credentials the user already filled in. The write uses
+ * `wx` (exclusive create) so the existence check and the create happen
+ * as a single syscall — there is no TOCTOU window where another process
+ * could create the file between a precheck and the write.
  */
 export function outputInitConfig(): void {
   const sourceUrl = new URL("../config.example.yaml", import.meta.url);
   const destPath = path.resolve("config.yaml");
 
-  if (fs.existsSync(destPath)) {
-    throw new Error(
-      `config.yaml already exists at ${destPath}. ` +
-        `Remove or rename it before running --init-config.`,
-    );
-  }
-
   const contents = fs.readFileSync(sourceUrl, "utf-8");
-  fs.writeFileSync(destPath, contents);
+  try {
+    fs.writeFileSync(destPath, contents, { flag: "wx" });
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === "EEXIST") {
+      throw new Error(
+        `config.yaml already exists at ${destPath}. ` +
+          `Remove or rename it before running --init-config.`,
+      );
+    }
+    throw err;
+  }
 
   const gitignoreNote = ensureGitignoreEntry(destPath);
 
@@ -171,7 +177,13 @@ async function main() {
     }
 
     if (cliOptions.initConfig) {
-      outputInitConfig();
+      try {
+        outputInitConfig();
+      } catch (error) {
+        const msg = error instanceof Error ? error.message : String(error);
+        console.error(`--init-config failed: ${msg}`);
+        process.exit(1);
+      }
       process.exit(0);
     }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ import {
   loadConfigFromYaml,
   MCPServerConfiguration,
 } from "@src/config/index.js";
+import { fs, path } from "@src/confluent/node-deps.js";
 import { TelemetryEvent, TelemetryService } from "@src/confluent/telemetry.js";
 import { ToolHandler } from "@src/confluent/tools/base-tools.js";
 import { ToolName } from "@src/confluent/tools/tool-name.js";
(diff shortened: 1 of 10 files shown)


src/index.ts
```diff
@@ -79,6 +80,79 @@ export function outputApiKey(): void {
...
+ * is the user's project but the example file lives next to the install.
+ *
+ * Refuses to overwrite an existing destination so an accidental rerun
+ * cannot wipe credentials the user already filled in. The write uses
+ * `wx` (exclusive create) so the existence check and the create happen
+ * as a single syscall — there is no TOCTOU window where another process
+ * could create the file between a precheck and the write.
+ */
+export function outputInitConfig(): void {
+  const sourceUrl = new URL("../config.example.yaml", import.meta.url);
```
jlrobins, line 100: For this release, when just one connection possible, we might want either two separate cmdline args (or a secondary one) to drive 'bootstrap me for ccloud oauth' vs 'bootstrap me for direct connections', which would just vary which config file they get cloned from here.

An easy followup, and then we could either decide to get rid of the difference when we could ship 'em a single integrated file with multiple connection blocks as the template, with comments saying 'pick at least one of these blocks.'

jlrobins (approved):
Good stuff. Comment about option for near future refinement, but this'll be great.

src/index.ts
flippingbits: Thanks, great idea! Tracking follow-up work here: https://github.com/confluentinc/mcp-confluent/issues/384

PR #359, opened 2026-05-05
Enable native Kafka tools under OAuth

bcscc:
## Summary of Changes

- Light up the OAUTHBEARER native-Kafka path for `list-topics`, `create-topics`, `delete-topics`, `produce-kafka-message`, and `consume-kafka-messages`. Each handler accepts `cluster_id` (`lkc-...`) and `environment_id` (`env-...`) tool args; under direct connections the args are ignored (cluster fixed by config), under OAuth they are required and the manager throws an actionable discovery hint when missing.
  - `OAuthClientManager` builds a fresh `KafkaJS.Kafka` per tool call, resolves the bootstrap endpoint via `cmk/v2/clusters/{id}` against the cloud REST client, and configures SASL/OAUTHBEARER through librdkafka's **synchronous** `oauthbearer_token_refresh_cb` (not the kafkaJS-compat    `sasl.oauthBearerProvider`). The compat layer's `async` provider returns a Promise that resolves on the microtask queue and races librdkafka's SASL state machine — using the native callback eliminates that race at the source, replacing the previous `try { listTopics(5s) } catch {
listTopics(30s) }` SASL-race retry with a single deterministic step (see next bullet).
  - Native Kafka clients are caller-owned and per-call: handlers wrap usage in `try { ... } finally { await disposeIfOAuth(runtime, connId, client) }`. `disposeIfOAuth` is a no-op on direct connections (`AsyncLazy` admin/producer singletons stay manager-owned) and `client.disconnect()`   on OAuth, swallowing + logging disconnect errors so disposal failure never masks the handler's own error. 
- One-shot `admin.listTopics()` metadata warmup remains inside OAuth admin construction (after `connect()`, before return). **Distinct from the SASL-race retry above**: even with SASL plumbed synchronously, `admin.connect()` returns before librdkafka has discovered the controller
broker, and `createTopics`/`deleteTopics` time out locally on first use without it. Cost: ~50–200ms per OAuth admin call.
- Extract `formatKafkaError(err)` alongside `resolveKafkaClusterArgs` and `disposeIfOAuth`. Centralizes Kafka-error rendering for `create-topics` (per-topic-cause unwrap from `KafkaJSAggregateError`), `produce-message`, and `consume-messages`. List/delete still let admin errors throw and propagate through the MCP layer.
[...]

PR #385, opened 2026-05-07
Refactor/357 predicate data member in all tools

### LINKED ISSUE #357: Handler layer: replace `enabledConnectionIds()` overrides with `readonly predicate` property

jlrobins:
## Handler `predicate` property

- Adds `abstract readonly predicate: ConnectionPredicate` to `BaseToolHandler` and makes `enabledConnectionIds()` and `connectionVerdicts()` concrete on the base — both methods do a single pass over `runtime.config.connections` against `this.predicate`, no extra indirection. Every handler now declares only its predicate; the coordination between predicate and the rest of the enabled-connection machinery lives in one place. 29 handler files transitioned: 26 leaf handlers each lost a 3-line override (most now read as a bare `readonly predicate = hasXxx;` line; the five native-Kafka leaves wrap with `widenForOAuth(hasKafkaBootstrap)` and the two compound leaves use `override readonly predicate = allOf(...)` — see next bullet), and the three domain bases (`FlinkToolHandler`, `TableflowToolHandler`, `ConnectToolHandler`) lost their predicate-bound override too.
- The two compound-predicate handlers from #356 (`CreateConnectorHandler`, `QueryProfilerHandler`) declare their predicates as `allOf(...)` rather than overriding methods. The unsafe raw-`&&` shape is now structurally unreachable: there's no method to override with bad boolean composition because the API surface is "name your predicate," not "compute your enabled-set."
- Adds `alwaysEnabled: ConnectionPredicate` to `connection-predicates.ts` for the no-domain-requirement case (`SearchProductDocsHandler`). Reads as exactly what it does at the use site: `readonly predicate = alwaysEnabled;`. Replaces the previous bespoke `Object.keys(runtime.config.connections)` branch.
- Adds `connectionVerdicts` to the `ToolHandler` interface so external consumers (the diagnostic-tool work in #358) can call it without a `BaseToolHandler` cast.

## Helper inlining + `@final` JSDoc

- Deleted `connectionIdsWhere` and `connectionReasonsWhere` from `connection-predicates.ts`. Both were single-consumer thin wrappers around language primitives — `Object.entries(…).filter(…).map(…)` in one, `new Map(Object.entries(…).map(…))` in the other — and their only callers were the matching `BaseToolHandler` methods. Inlined into those methods directly; the seven helper-isolation tests went with the helpers (3 for `connectionIdsWhere`, 4 for `connectionReasonsWhere`). Predicate-derived behavior stays covered by 4 cases in `base-tools.test.ts` plus the 43 handler unit-test files that already exercise `enabledConnectionIds(runtime)` end-to-end. The `connection-predicates.ts` API surface is now strictly the predicate library: predicates, `allOf`, `alwaysEnabled`. No iteration helpers.
- Marked `enabledConnectionIds` and `connectionVerdicts` `@final` in JSDoc with a one-line "subclasses must not override" rationale. TypeScript has no native `final` keyword; the tag is documentation-grade enforcement layered on top of the structural mitigation — subclasses have no reason to override because the `predicate` property is the customization seam, and overriding the methods would let `enabledConnectionIds` and `connectionVerdicts` disagree about which connections a tool is enabled on.

## Grouped startup logging

- Replaces the per-tool "Tool X disabled; no connections satisfy its requirements" warn with a grouped warn keyed on `(connectionId, reason)`. New shape: `Tools disabled on connection 'default' — no 'flink' block in connection config: list-flink-statements, create-flink-statement, …`. One log line per missing config piece, listing every tool blocked by it, instead of N copies of "X disabled" with no diagnostic content. Per-tool _enabled_ info logs and the allow/block-list per-tool warns stay as-is — those reasons aren't predicate-driven.
- Introduces `groupDisabledToolsByReason(handlers, runtime): DisabledToolGroup[]` in a new `tool-availability.ts` module. Pure function — takes an iterable of `(ToolName, ToolHandler)` pairs and a runtime, returns groups ordered lexicographically by `connectionId` (deterministic, engine-independent — JS plain-object iteration order is not a portable invariant for `Record<string, …>` because integer-like keys get reordered numerically). A tool is included only when it's "fully disabled" (every connection's verdict is non-enabled); tools enabled on at least one connection are omitted entirely so users don't see noise about tools that are actually working. Unit-tested with seven cases covering empty input, all-enabled, single-disabled, same-reason collapse, different-reasons split, the partially-enabled omission rule, and lexicographic-order pinning across connections inserted in non-alphabetical order.
- Issue #357 named `ServerRuntime.fromConfig()` as the host for the new logging, but `ServerRuntime` doesn't currently know about tools or the registry. Logging stayed in `index.ts`'s `getToolHandlersToRegister()` where both runtime and registry are already at hand, keeping `ServerRuntime` tool-agnostic. Confirmed with James before doing it this way.

## Composition with `widenForOAuth` (post-merge with #359)

- After merging #359, the five native-Kafka handlers (`consume`, `create-topics`, `delete-topics`, `list-topics`, `produce`) declare `readonly predicate = widenForOAuth(hasKafkaBootstrap);`. The combinator from #359 plugs into the property-API with no special-casing — `widenForOAuth` is just another `(predicate) => predicate` factory, and the base class derives `enabledConnectionIds()` and `connectionVerdicts()` from whatever predicate the handler names. The five OAuth-typed `enabledConnectionIds()` test cases #359 added (one per kafka handler) stay green against the property-API implementation, confirming the seam composes correctly.
- The file-header OAuth note in `connection-predicates.ts` was tightened to reference `widenForOAuth` as the canonical mechanism for OAuth-capable handlers (alongside `hasConfluentCloud`'s built-in OAuth pass), replacing the "predicates widen on a case-by-case basis" hand-wave.

### Try it locally

After `npm run build`, point the server at any fixture in `test-fixtures/yaml_configs/valid/` to see the new grouped-by-reason warns:

```bash
node dist/index.js -c <fixture-path> </dev/null 2>&1 \
  | grep -E '"Tools disabled|tool\(s\) enabled"'
```

The `</dev/null` closes stdio so the server shuts down once startup is done; the `grep` keeps just the grouped warns plus the final "N tool(s) enabled" summary.

Sample fixtures to try, each producing a different mix of enabled vs grouped-disabled output:

- `valid/kafka-only.yaml` — six tools enabled (native-Kafka group + `search-product-docs`); one grouped warn per missing-block reason for flink/ccloud/sr/tableflow/telemetry, plus `'kafka' block does not have 'rest_endpoint' field` covering the two kafka REST tools.
- `valid/flink-with-required-fields.yaml` — only flink + `search-product-docs` enabled; every other domain grouped under its own missing-block reason.
- `valid/ccloud-and-flink-with-auth.yaml` — flink + the ccloud REST tools enabled; kafka/sr/tableflow/telemetry grouped.
- A two-connection custom fixture (drop one in `/tmp/`) to see two parallel grouped warns, each keyed on its own `connectionId` — confirms the `(connectionId, reason)` grouping isn't collapsing across connections.

or human-readable output rather than JSON envelopes, prefilter to JSON lines and pipe through `jq`:

```bash
node dist/index.js -c <fixture-path> </dev/null 2>&1 \
  | grep '^{' \
  | jq -r 'select(.msg | startswith("Tools disabled")) | .msg'
```

For example:
```bash
$ node dist/index.js -c test-fixtures/yaml_configs/valid/ccloud-with-auth.yaml </dev/null 2>&1   | grep '^{' | jq -r 'select(.msg | startswith
("Tools disabled")) | .msg'
Tools disabled on connection 'production' — no 'kafka' block in connection config: list-topics, create-topics, delete-topics, produce-message, consume-messages, create-connector, alter-topic-config, get-topic-config
Tools disabled on connection 'production' — no 'flink' block in connection config: list-flink-statements, create-flink-statement, read-flink-statement, delete-flink-statements, get-flink-statement-exceptions, list-flink-catalogs, list-flink-databases, list-flink-tables, describe-flink-table, get-flink-table-info, check-flink-statement-health, detect-flink-statement-issues, get-flink-statement-profile
Tools disabled on connection 'production' — no 'schema_registry' block in connection config: search-topics-by-tag, search-topics-by-name, create-topic-tags, delete-tag, remove-tag-from-entity, add-tags-to-topic, list-tags, list-schemas, delete-schema
Tools disabled on connection 'production' — no 'tableflow' block in connection config: create-tableflow-topic, list-tableflow-regions, list-tableflow-topics, read-tableflow-topic, update-tableflow-topic, delete-tableflow-topic, create-tableflow-catalog-integration, list-tableflow-catalog-integrations, read-tableflow-catalog-integration, update-tableflow-catalog-integration, delete-tableflow-catalog-integration
```

## Test coverage

- Existing handler tests assert on `enabledConnectionIds(runtime)` outcomes — those keep passing untouched, confirming the refactor preserved behavior across all 26 leaves and 3 domain bases.
- New `BaseToolHandler` tests pin the predicate-derived behavior directly: `enabledConnectionIds()` and `connectionVerdicts()` against an enabled stub and a disabled stub.
- New `groupDisabledToolsByReason()` tests cover the seven cases above; new `getToolHandlersToRegister` test asserts the exact grouped-warn message format (collapses two same-reason tools into one log line).

### Followup: per-handler test slim-down (three-PR sequence, out of scope here)

With `predicate` as the sole customization seam and `enabledConnectionIds`/`connectionVerdicts` derived on the base class, the per-handler `enabledConnectionIds(runtime)` cases now duplicate coverage that already lives at two stronger levels: `base-tools.test.ts` pins the derivation exactly once, and `connection-predicates.test.ts` pins each predicate's verdict exhaustively. What remains for a per-handler test to prove is just "the right predicate is wired."

Leaving the existing handler tests in place rather than churn the ~43 unit test files alongside the refactor itself. The followup sequence (three PRs total, issues filed):

1. **Predicate library completion (#389, one PR).** Promote each unique composed predicate to a named export in `connection-predicates.ts` with a one-line intent docstring and focused predicate-level tests. Concrete examples: `widenForOAuth(hasKafkaBootstrap)` (shared by the five native-Kafka handlers from #359) becomes a named const; `allOf(...)` compositions used by `CreateConnectorHandler` and `QueryProfilerHandler` (from #356) get named too. After this PR every handler's `predicate` property is a bare named-export reference.
2. **Handler test slim-down (#390, one mechanical PR sweeping all 13 domains).** Replace each handler's `describe("enabledConnectionIds()")` block with a single `expect(handler.predicate).toBe(theNamedPredicate)` identity assertion across `billing, catalog, clusters, connect, docs, environments, flink, kafka, metrics, organizations, schema, search, tableflow`. `getToolConfig()` and `handle()` tests stay untouched. The work is uniformly mechanical — the same edit pattern repeated across every handler test file, the redundancy itself the asset — so a single PR lets a reviewer verify the pattern once and `grep` the diff to confirm ~26 identical repetitions. Splitting into 13 per-domain PRs would add review-cycle overhead with no per-PR insight value.
3. **Cleanup pass (#391, one PR).** Audit `tests/factories/runtime.ts` for fixture factories whose only remaining consumers were the deleted handler tests; fold or delete. Re-ask whether `enabledConnectionIds` and `connectionVerdicts` still need to be on the public `ToolHandler` interface — startup-logging and the `describe-tool-availability` consumer (#358) probably keep them load-bearing, but worth checking once the per-handler call sites are gone.

The deliberate choice here is **identity-based wiring tests across the board** rather than a mix of identity (for bare predicates) and behavioral (for composed ones). Promoting compositions to named exports makes the predicate library a complete catalog of "valid tool gates," reads better at the use site than re-decoding `widenForOAuth(hasKafkaBootstrap)` inline, and keeps the per-handler test shape uniform.

## Relationships

- Closes #357
- Builds on #356 (predicate-verdict types + `allOf` combinator)
- Unblocks #358 (the `describe-tool-availability` MCP tool) — that PR can now call `handler.connectionVerdicts(runtime)` from any consumer that holds a `ToolHandler` reference.
- Spawns #389 → #390 → #391 — the three-PR followup chain detailed under `## Test coverage` above.

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes

### DIFF
--- a/src/confluent/tools/connection-predicates.ts
+++ b/src/confluent/tools/connection-predicates.ts
@@ -4,12 +4,11 @@
 // so callers can explain why a tool is absent rather than silently dropping
 // it from the catalogue.
 //
-// Tool handlers reach this module via `connectionIdsWhere` from their
-// `enabledConnectionIds()` method — the returned id list drives whether MCP
-// advertises the tool to its clients. `connectionReasonsWhere` returns the
-// full per-connection verdict map, the canonical shape for diagnostics that
-// need to render the verdict (group disabled tools by reason, surface
-// availability per connection) rather than just filter on it.
+// Tool handlers consume predicates indirectly: each handler declares a
+// `predicate` property, and `BaseToolHandler` derives both
+// `enabledConnectionIds()` and `connectionVerdicts()` from it. Compose
+// compound requirements with `allOf(...)` and use `alwaysEnabled` for
+// tools with no service-block requirement.
 //
 // OAuth note: while OAuth support is being built out, most predicates
 // currently short-circuit on `conn.type === "oauth"` and answer disabled
@@ -200,11 +199,18 @@ export function hasCCloudCatalogSupport(
 }
 
 /**
+ * Predicate that returns {@linkcode ENABLED} for every connection. Use as
+ * the {@linkcode BaseToolHandler.predicate} for tools with no service-block
+ * requirement (e.g., generic docs search).
+ */
+export const alwaysEnabled: ConnectionPredicate = () => ENABLED;
+
+/**
  * Combine predicates with logical AND, short-circuiting on the first
  * failure. Returns {@linkcode ENABLED} only when every predicate passes for
  * the given connection; otherwise returns the first failing verdict so the
- * specific reason propagates to {@linkcode connectionReasonsWhere} and
- * downstream diagnostics.
+ * specific reason propagates downstream to startup logging and the
+ * diagnostic-tool surface.
  *
  * Use this — never raw `predA(conn) && predB(conn)`. JavaScript boolean
  * composition silently drops the first operand because every
@@ -220,29 +226,6 @@ export function allOf(
     }
     return ENABLED;
   };
-}
-
-export function connectionIdsWhere(
-  connections: Readonly<Record<string, ConnectionConfig>>,
-  predicate: ConnectionPredicate,
-): string[] {
-  return Object.entries(connections)
-    .filter(([, conn]) => predicate(conn).enabled)
-    .map(([id]) => id);
-}
-
-/**
- * Returns the full predicate verdict for every connection, preserving
- * insertion order — the canonical shape for diagnostics that need to group
- * disabled connections by reason or render per-connection availability.
- */
-export function connectionReasonsWhere(
-  connections: Readonly<Record<string, ConnectionConfig>>,
-  predicate: ConnectionPredicate,
-): Map<string, PredicateResult> {
-  return new Map(
-    Object.entries(connections).map(([id, conn]) => [id, predicate(conn)]),
-  );
 }
 
 /**
--- a/src/confluent/tools/tool-availability.test.ts
+++ b/src/confluent/tools/tool-availability.test.ts
@@ -1,0 +1,174 @@
+import { ToolHandler } from "@src/confluent/tools/base-tools.js";
+import {
+  ConnectionPredicate,
+  ToolDisabledReason,
+  alwaysEnabled,
+} from "@src/confluent/tools/connection-predicates.js";
+import { groupDisabledToolsByReason } from "@src/confluent/tools/tool-availability.js";
+import { ToolName } from "@src/confluent/tools/tool-name.js";
+import {
+  KAFKA_CONN,
+  bareRuntime,
+  runtimeWith,
+} from "@tests/factories/runtime.js";
+import { StubHandler } from "@tests/stubs/index.js";
+import { describe, expect, it } from "vitest";
+
+function stubWithPredicate(predicate: ConnectionPredicate): ToolHandler {
+  const handler = new StubHandler();
+  // Override the readonly predicate via type assertion for test composition.
+  (handler as unknown as { predicate: ConnectionPredicate }).predicate =
+    predicate;
+  return handler;
+}
+
+const disabledForKafka: ConnectionPredicate = () => ({
+  enabled: false,
+  reason: ToolDisabledReason.MissingKafkaBlock,
+});
+
+const disabledForFlink: ConnectionPredicate = () => ({
+  enabled: false,
+  reason: ToolDisabledReason.MissingFlinkBlock,
+});
+
+describe("tool-availability.ts", () => {
+  describe("groupDisabledToolsByReason()", () => {
+    it("should return an empty array when no tools are passed", () => {
+      expect(groupDisabledToolsByReason([], bareRuntime())).toEqual([]);
+    });
+
+    it("should return an empty array when every tool is enabled on every connection", () => {
+      const handler = stubWithPredicate(alwaysEnabled);
+      expect(
+        groupDisabledToolsByReason(
+          [[ToolName.LIST_TOPICS, handler]],
+          runtimeWith(KAFKA_CONN),
+        ),
+      ).toEqual([]);
+    });
+
+    it("should produce one group per (connectionId, reason) when one tool is disabled", () => {
+      const handler = stubWithPredicate(disabledForKafka);
+      const groups = groupDisabledToolsByReason(
+        [[ToolName.LIST_TOPICS, handler]],
+        runtimeWith({ schema_registry: { endpoint: "http://sr" } }),
+      );
+      expect(groups).toEqual([
+        {
+          connectionId: "default",
+          reason: ToolDisabledReason.MissingKafkaBlock,
+          toolNames: [ToolName.LIST_TOPICS],
+        },
+      ]);
+    });
+
+    it("should collapse multiple tools failing for the same (connectionId, reason) into one group", () => {
+      const handlerA = stubWithPredicate(disabledForKafka);
+      const handlerB = stubWithPredicate(disabledForKafka);
+      const groups = groupDisabledToolsByReason(
+        [
+          [ToolName.LIST_TOPICS, handlerA],
+          [ToolName.CREATE_TOPICS, handlerB],
+        ],
+        runtimeWith({ schema_registry: { endpoint: "http://sr" } }),
+      );
+      expect(groups).toEqual([
+        {
+          connectionId: "default",
+          reason: ToolDisabledReason.MissingKafkaBlock,
+          toolNames: [ToolName.LIST_TOPICS, ToolName.CREATE_TOPICS],
+        },
+      ]);
+    });
+
+    it("should split into separate groups when tools fail with different reasons", () => {
+      const kafkaTool = stubWithPredicate(disabledForKafka);
+      const flinkTool = stubWithPredicate(disabledForFlink);
+      const groups = groupDisabledToolsByReason(
+        [
+          [ToolName.LIST_TOPICS, kafkaTool],
+          [ToolName.LIST_FLINK_STATEMENTS, flinkTool],
+        ],
+        runtimeWith({ schema_registry: { endpoint: "http://sr" } }),
+      );
+      expect(groups).toEqual([
+        {
+          connectionId: "default",
+          reason: ToolDisabledReason.MissingKafkaBlock,
+          toolNames: [ToolName.LIST_TOPICS],
+        },
+        {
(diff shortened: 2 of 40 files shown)


src/confluent/tools/connection-predicates.ts
```diff
@@ -199,12 +199,19 @@ export function hasCCloudCatalogSupport(
   return ENABLED;
 }
 
+/**
+ * Predicate that returns {@linkcode ENABLED} for every connection. Use as
+ * the {@linkcode BaseToolHandler.predicate} for tools with no service-block
+ * requirement (e.g., generic docs search).
+ */
+export const alwaysEnabled: ConnectionPredicate = () => ENABLED;
```
shouples, line 207: Oh nice, I was going to add this as part of the upcoming connection-centric tools - makes that branch smaller!

src/confluent/tools/tool-availability.test.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    it("should omit tools that are enabled on at least one connection (only fully-disabled tools group)", () => {
+      // Predicate enables only on connections carrying a kafka block.
+      const partiallyEnabled: ConnectionPredicate = (conn) =>
+        conn.type === "direct" && conn.kafka !== undefined
+          ? { enabled: true }
+          : {
+              enabled: false,
+              reason: ToolDisabledReason.MissingKafkaBlock,
+            };
```
shouples, line 117: Isn't this the same as the `hasKafka` ConnectionPredicate?

src/confluent/tools/tool-availability.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      // grouping helper's read path.
+      (
+        runtime.config.connections as Record<
+          string,
+          (typeof runtime.config.connections)[string]
+        >
+      )["other"] = {
+        type: "direct",
+        schema_registry: { endpoint: "http://sr" },
+      };
```
shouples, line 132: Readability for these overrides are rough, could use a helper function

shouples (approved):
Thanks for the great cleanup here! Just one question and a minor cleanup suggestion, non-blocking though

src/confluent/tools/connection-predicates.ts
jlrobins: you had good feeling that we should go this way a while ago. Now we're there!

src/confluent/tools/tool-availability.test.ts
jlrobins: Will research here in the next branch, thanks.

PR #393, opened 2026-05-07
Add --init-oauth-config flag to bootstrap a starter OAuth config.yaml

### LINKED ISSUE #384: Add --init-oauth-config flag to bootstrap a starter OAuth config.yaml
**Follow-up to #378 / #379.** Same audience (`npx`-only users with no checkout), but for users who want OAuth auth against Confluent Cloud rather than API keys.

### Problem

`--init-config` writes a copy of `config.example.yaml`, which is heavily annotated for direct (api-key) connections — every block has `key:` / `secret:` placeholders, `${VAR}` interpolation hints, and per-service tuning. For users who want OAuth (added recently as `type: oauth` in `src/config/models.ts:40-43`), almost none of that applies: the OAuth connection has no service blocks, no API keys, and an optional `ccloud_env` that defaults to `prod`. The minimal OAuth YAML is three lines (see `oauth.yaml` in this repo).

So an OAuth user running `--init-config` today gets a config that is mostly noise to delete. They have to either know which lines to strip or hand-write the OAuth config from the docs. The bootstrap-without-repo win from #379 doesn't extend to them.

### Proposed design

Add a sibling early-exit flag to `--init-config`:

```bash
npx @confluentinc/mcp-confluent --init-oauth-config
# wrote ./config.yaml (added to .gitignore; next: edit and run with --config ./config.yaml)
```

Behavior is identical to `--init-config` — same destination (`./config.yaml`), same EEXIST refusal, same idempotent `.gitignore` append, same early-exit position in `src/index.ts:179`. The only differences:

1. The source template is a new `config.oauth.example.yaml` shipped in the npm tarball (added to `package.json#files` alongside `config.example.yaml`).
2. The success message points OAuth users at `--config ./config.yaml` (no extra `--oauth` flag — the YAML already declares `type: oauth`).

Wired into `src/cli.ts` as a new `CLIOptions.initOauthConfig` field, handled in `src/index.ts` adjacent to the existing `cliOptions.initConfig` branch. `outputInitConfig()` becomes parameterized over the source-file URL; the two flags share `ensureGitignoreEntry()` verbatim.

`--init-config` and `--init-oauth-config` are mutually exclusive (passing both should error out at parse time, like other CLI conflict checks in `assertNoConfigConflicts`).

### What goes in `config.oauth.example.yaml`

Roughly this (final wording up to the PR):

```yaml
# Example MCP Confluent Server configuration — OAuth (PKCE) connection.
#
# Generate `config.yaml` from this template with:
#   npx @confluentinc/mcp-confluent --init-oauth-config
#
# OAuth authenticates against Confluent Cloud via PKCE; no API keys to
# provision. The browser flow runs on first tool call. OAuth-eligible tools
# auto-enable — there are no per-service blocks like there are for direct
# (api-key) connections. Resource IDs (cluster_id, env_id, etc.) are passed
# as tool arguments at call time rather than declared here.

# --- Server ---
# (Same shape as config.example.yaml — uncomment if you need HTTP/SSE.)
# server:
#   transports: [stdio]
#   ...

connections:
  ccloud-oauth:
    type: oauth
```

Two things worth calling out:

- The `server:` block is intentionally commented-out (not omitted). OAuth users running over HTTP/SSE still need transport/auth/host settings, and the friendliest answer is to have the example show them where to put it without making stdio users delete a block.
- No `${VAR}` placeholders. OAuth connections have no secrets in the YAML, so interpolation has nothing to interpolate. (Server-side `server.auth.api_key` is the one secret an OAuth user might add — but it's commented out and shared with `config.example.yaml` anyway.)

### `.gitignore` behavior

Same as `--init-config`. Even though the bare OAuth YAML has no secrets, two cases justify keeping the gitignore step:

1. Users who add `server.auth.api_key` to gate their HTTP transport.
2. Consistency: one `--init-*` flag that touches `.gitignore` and another that doesn't would be a confusing surface.

Idempotence carries over for free: if a user runs both flags in sequence (the second one will EEXIST-refuse, but if they `rm config.yaml` between them) the gitignore append already de-dupes.

### Open design questions

- **Filename** — `config.oauth.example.yaml` vs. `config.example.oauth.yaml` (no strong opinion; current repo has just `config.example.yaml`).
- **Destination** — keep `./config.yaml` (current proposal) vs. `./config.oauth.yaml` so both `--init-*` flags can coexist in one directory.
- **Mutex** — should `--init-config` and `--init-oauth-config` actually be mutually exclusive, or just last-one-wins?

### Out of scope

- A `--init-config --oauth` subflag form instead of a separate `--init-oauth-config`. Symmetric to `--oauth` on the server-start path, but adds a parse-time mode switch to a flag whose current behavior is "copy one specific file." A separate flag is more discoverable in `--help` and easier to document.
- Schema-derived generation of the OAuth example from `oauthConnectionSchema` in `src/config/models.ts:410`. Same trade-off as #378 — the schema is two fields, drift risk is negligible, hand-maintained file wins on annotation quality.
- A combined flag that picks template based on a separate `--oauth` switch. Users running `--init-*` haven't decided on transport yet; coupling the bootstrap step to `--oauth` (which today *also* implies "and run the server now") muddies both flags.

### Context

- Sibling flag for reference: `--init-config` (`src/cli.ts:214-217`, handler at `src/index.ts:99-122`, gitignore helper at `src/index.ts:133-154`).
- OAuth connection schema: `src/config/models.ts:410-415` (just `type` + optional `ccloud_env`).
- OAuth eligibility widening that auto-enables tools without explicit blocks: `connection-predicates.ts` (`hasConfluentCloud` returns true for OAuth connections).
- Predecessor PR: #379. Predecessor issue: #378.

flippingbits:
## Summary

- Adds `--init-oauth-config` as a sibling early-exit flag to `--init-config`. Drops a copy of the new bundled `config.oauth.example.yaml` template (minimal `connections.<name>: { type: oauth }`) into `./config.yaml`, idempotently appends to `.gitignore`, refuses to overwrite an existing file, and exits.
- `outputInitConfig(oauth?)` is parameterized over a single boolean that selects both the bundled template and the flag name in the EEXIST error; defaults to `false` to preserve the existing call shape.
- The two `--init-*` flags are mutually exclusive at parse time, matching the conflict-check style already in `assertNoConfigConflicts`.

Fixes #384


src/index.ts
jlrobins: No credentials if oauth. Conditionalize?

jlrobins (approved):
Works!

SC complaining at the complexity creep in main() and parsing cli options. Not quite new here, just 'now a little worse'.

Could maybe collect all of the 'early exits' out of main and into a 'handleEarlyExits(CLIOptions)' or something? Then also easy gains over getting more test coverage there also.

src/index.ts
flippingbits: Fixed in 42d58fd

jlrobins (approved):
Thanks!

Issue #390, opened 2026-05-07
Per-handler test slim-down: wiring-only assertions across all handlers


jlrobins:
After #357 + the predicate library completion followup, per-handler `enabledConnectionIds(runtime)` cases duplicate coverage that already lives at two stronger levels:

- `base-tools.test.ts` pins the derivation (predicate → enabled-set) exactly once.
- `connection-predicates.test.ts` pins each predicate's verdict exhaustively against the canonical connection fixtures.

What remains for a per-handler test to prove is just **"the right predicate is wired."** Everything else is redundant.

## Scope (single mechanical PR across all 13 handler domains)

For every handler test file in `src/confluent/tools/handlers/{billing,catalog,clusters,connect,docs,environments,flink,kafka,metrics,organizations,schema,search,tableflow}/`:

Replace the existing `describe("enabledConnectionIds()")` block (typically 3–5 `it()` cases) with a single identity assertion:

```typescript
it("should declare the expected predicate", () => {
  expect(handler.predicate).toBe(theNamedPredicate);
});
```

Drop now-unused `runtime` factory imports at the use sites (`bareRuntime`, `kafkaRuntime`, `flinkRuntime`, `ccloudOAuthRuntime`, etc.). `getToolConfig()` and `handle()` tests stay untouched.

## Why a single PR rather than 13

The work is **uniformly mechanical** — the same edit pattern repeated across every handler test file. Splitting into 13 per-domain PRs adds review-cycle overhead with no per-PR insight value:

- A reviewer needs to see the pattern _once_ and verify it. After that, every additional handler is a redundant repeat that the diff itself proves out (every `it("should declare the expected predicate")` line is structurally identical).
- The redundancy is the **asset**: a reviewer can grep the diff for that one assertion shape and confirm ~26 copies. Anything that doesn't match that shape is suspect; the per-PR review would be running the same grep with extra steps.
- Conflict risk is negligible — nothing else is concurrently editing these test blocks.
- One rebase if the PR sits a few days; thirteen rebases otherwise.

The only cost (a bigger diff) is paid once and recovered immediately at review time.

## Acceptance

- Every handler test file's `describe("enabledConnectionIds()")` block contains exactly one `expect(handler.predicate).toBe(...)` assertion.
- Total `it()` count across handler test files drops by ~3–5× (depending on how many cases each old block had).
- No `enabledConnectionIds(runtime)` calls remain in handler test files.
[...]

PR #395, opened 2026-05-07
Lockdown tool predicates to known (tested) set.

### LINKED ISSUE #389: Predicate library completion: promote composed predicates to named exports
After #357 lands, every tool handler declares a `predicate` property. Most are bare references (`readonly predicate = hasKafka;`), but a handful are inline compositions:

- `widenForOAuth(hasKafkaBootstrap)` — used by the five native-Kafka handlers (`ConsumeKafkaMessagesHandler`, `CreateTopicsHandler`, `DeleteTopicsHandler`, `ListTopicsHandler`, `ProduceKafkaMessageHandler`)
- `allOf(hasDirectConfluentCloud, hasKafkaAuth)` — `CreateConnectorHandler`
- `allOf(hasFlink, hasTelemetry)` — `QueryProfilerHandler`

## Why promote them

**Intent at the use site.** `readonly predicate = oauthCapableNativeKafka;` reads better than re-decoding `widenForOAuth(hasKafkaBootstrap)` in your head every time you land in a handler. The composition's _purpose_ is what the next reader needs; the _mechanics_ are an implementation detail of the predicate library.

**Uniform handler-test wiring.** Sets up the next followup (per-handler test slim-down) where every handler's `enabledConnectionIds()` test becomes `expect(handler.predicate).toBe(theNamedPredicate)`. Inline compositions can't be identity-checked because each call returns a fresh closure; named exports give a stable reference for `toBe`.

**Catalog completeness.** Today `connection-predicates.ts` exports the kit-of-parts (`hasKafka`, `widenForOAuth`, `allOf`). After this work it also exports the assembled vocabularies callers actually want (`oauthCapableNativeKafka`, `directConnectorRequirement`, `flinkProfilerRequirement` — names TBD). The library becomes a complete catalog of "valid tool gates" rather than just primitives plus glue.

## Scope

For each unique composition currently used by a handler:

1. Add a named `const` export to `src/confluent/tools/connection-predicates.ts` with a one-line intent docstring (focus on _what gate this represents_, not _how it's composed_).
2. Add focused predicate-level tests in `connection-predicates.test.ts` matching the depth of the existing `hasKafka` / `hasFlink` tests — exhaustive against the canonical `ConnectionConfig` fixtures (direct connections enabled, direct connections disabled, OAuth connection).
3. Update handler use sites to reference the named export.

Naming is the hard part — pick names at the level of "what tool gate is this," not the predicate's compositional structure. The PR author owns naming; existing predicates in the file are the closest model.

## Acceptance

- Every handler's `predicate` property is a bare named-export reference. No `widenForOAuth(...)` or `allOf(...)` at handler use sites.
- Every named composed predicate has its own predicate-level test block in `connection-predicates.test.ts`.
- `npm run typecheck`, `npm run lint`, and the unit suite all green.

## Relationships

- Child of #357 (which established the `predicate`-property API)
- Blocks the per-handler test slim-down followup

jlrobins:
## Predicate library completion

## TL;DR ##

Make every actual handler predicate a known named and tested function so that we can then stop testing `getConnectionIds()` at the per-tool-level, since it is now singularly implemted in `BaseToolHandler` atop the concrete tool's `predicate`.

## TL ##

Three composed predicates were inlined at handler use sites after #357 landed. This PR promotes each to a named `const` export in `connection-predicates.ts`, swaps the seven handler use sites to reference the named exports, and publishes the assembled vocabularies callers actually want — not just the kit-of-parts.

| Inline composition                             | Named export               | What it gates                                                                                                                                                                                                                                                                    |
| ---------------------------------------------- | -------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `widenForOAuth(hasKafkaBootstrap)`             | `kafkaBootstrapOrOAuth`    | Native-Kafka client gate, widened to admit OAuth. Direct connections still need `kafka.bootstrap_servers`; OAuth satisfies it unconditionally (broker URL synthesized from the Auth0 environment). Used by 5 native-Kafka handlers.                                              |
| `allOf(hasDirectConfluentCloud, hasKafkaAuth)` | `canCreateDirectConnector` | Requires both `confluent_cloud` (the `/connect/v1` endpoint) and `kafka.auth` (the connector spec carries kafka API credentials). Strict-direct — OAuth answers `OAuthNotDirectCapable` because the handler calls `getSoleDirectConnection()`. Used by `CreateConnectorHandler`. |
| `allOf(hasFlink, hasTelemetry)`                | `flinkWithTelemetry`       | Requires both blocks — the tool reads telemetry metrics about Flink statements. Used by `QueryProfilerHandler`.                                                                                                                                                                  |

The `<basePredicate>OrOAuth` shape on the first row is intended as a scaling convention: as more handlers are brought into the OAuth fold, the suffix tells a reader at a glance "this gate has been widened — here's the base it widens." The name avoids overstating with terms like "capable" or "client" — the disjunction is exactly what the predicate implements.

## Predicate-level tests

A new `describe` block per named composite in `connection-predicates.test.ts`, each modeled on the existing `hasKafka` / `hasFlink` shape:

- `kafkaBootstrapOrOAuth` — direct enabled (`KAFKA_CONN`), direct disabled with `MissingKafkaBootstrap` (`KAFKA_REST_CONN`), direct disabled with `MissingKafkaBlock` (`SCHEMA_REGISTRY_CONN`), OAuth enabled (proves the widening).
- `canCreateDirectConnector` — direct enabled (new `DIRECT_CCLOUD_KAFKA_AUTH_CONN` fixture), direct disabled with `MissingConfluentCloudBlock` via first-conjunct fail, direct disabled with `MissingKafkaBlock` via second-conjunct fail, OAuth disabled with `OAuthNotDirectCapable`.
- `flinkWithTelemetry` — direct enabled (new `FLINK_AND_TELEMETRY_CONN` fixture), direct disabled with `MissingFlinkBlock` via first-conjunct fail, direct disabled with `MissingTelemetryBlock` via second-conjunct fail, OAuth disabled with `OAuthNoServiceBlocks`.

Twelve new test cases total, fail/pass reasons matching what the underlying combinators produce — the same depth and shape as the rest of `connection-predicates.test.ts`.

## Use-site swaps

Seven handlers updated:

- 5 native-Kafka leaves (`consume-kafka-messages-handler.ts`, `create-topics-handler.ts`, `delete-topics-handler.ts`, `list-topics-handler.ts`, `produce-kafka-message-handler.ts`) — `readonly predicate = kafkaBootstrapOrOAuth;`
- `connect/create-connector-handler.ts` — `override readonly predicate = canCreateDirectConnector;` (preserves `override` since the leaf supersedes `ConnectToolHandler`'s domain predicate). The handler's stale internal comment that referenced `hasKafkaAuth in enabledConnectionIds()` was retargeted to the new predicate name.
- `flink/diagnostics/query-profiler-handler.ts` — `override readonly predicate = flinkWithTelemetry;`

After this PR every handler's `predicate` property is a bare named-export reference. No `widenForOAuth(...)` or `allOf(...)` at handler use sites.

## Documentation updates

The PR establishes a hard rule across three docs: the `predicate` property must reference a named export from `connection-predicates.ts`. No inline `allOf(...)` or `widenForOAuth(...)` at handler use sites. When no existing export fits, the path is "add a named `const` export → add a per-predicate test → reference the new export"; promotion is a precondition of landing the handler, not a followup. The rule's enforcement is the new `predicate property` block in `tool-registry.test.ts` (see "Whole-tool-spanning enforcement" below) — inline composition produces a fresh closure that fails the membership check, with the offending tool name in the row label.

- The `predicate` docstring on `BaseToolHandler` ([base-tools.ts](src/confluent/tools/base-tools.ts)) leads with three illustrative one-liners (`hasKafka`, `kafkaBootstrapOrOAuth`, `alwaysEnabled`) and states the rule directly.
- [.claude/rules/tool-handlers.md](.claude/rules/tool-handlers.md) carries the same rule under "Implement these members" and re-frames the "Connection Predicates" bullets so `allOf` / `widenForOAuth` are documented as construction tools used inside `connection-predicates.ts`, not handler-side glue.
- The top-of-file comment in [connection-predicates.ts](src/confluent/tools/connection-predicates.ts) is updated to match — handlers consume named exports; combinators stay inside the library.

## Whole-tool-spanning enforcement

A new `describe("predicate property")` block in [tool-registry.test.ts](src/confluent/tools/tool-registry.test.ts) makes the "must be a named export" rule executable rather than aspirational. It maintains an explicit `NAMED_PREDICATES` allow-list typed as `ReadonlySet<ConnectionPredicate>` and asserts per tool (via `it.each(ALL_TOOL_NAMES)`) that `handler.predicate` is a member. The instanceof narrowing to `BaseToolHandler` runs first — that's the class where `predicate` lives — and is itself a useful invariant.

The type annotation is the load-bearing piece. `allOf` has signature `(...preds: ConnectionPredicate[]) => ConnectionPredicate` and `widenForOAuth` has signature `(p: ConnectionPredicate) => ConnectionPredicate`; neither is assignable to `(conn: ConnectionConfig) => PredicateResult`. So the combinators (and any future helper that returns a predicate rather than is one) cannot be added to the set even by mistake.

Failure modes the test catches:

- **Inline composition at a handler use site.** `readonly predicate = allOf(...)` returns a fresh closure not in the allow-list; the per-tool row fails with the offending tool name in the label.
- **Allow-list drift after adding a new predicate.** Adding a named predicate to `connection-predicates.ts` without adding it to `NAMED_PREDICATES` causes loud per-handler failures for every tool that uses the new predicate. The assertion message distinguishes the two causes so a maintainer reading the failure can act without diagnosis.

Red-step verified: temporarily flipping `list-topics-handler.ts` back to `widenForOAuth(hasKafkaBootstrap)` caused exactly that one row to fail with the documented remediation in the assertion message.

## Behavior preservation

The OAuth tool-surface partition test in [src/index.test.ts](src/index.test.ts) (`describe("against configured OAuth connection")`) runs through `getToolHandlersToRegister(ccloudOAuthRuntime())` and asserts the aggregate enabled set against `EXPECTED_OAUTH_ENABLED`. It's predicate-spelling-agnostic: any name swap that accidentally widened or narrowed a gate would surface there. Combined with the per-predicate unit tests and the new whole-tool-spanning enforcement, behavioral drift has nowhere to hide. Suite stays green at 1332 tests (was 1265 pre-PR; +12 from the new predicate-test cases, +55 from the per-handler `predicate property` block plus its two sanity tests).

## Relationships

- Closes #389
- Child of #357 (which established the `predicate`-property API)
- Unblocks #390 (per-handler test slim-down) — every handler's `predicate` is now a bare named-export reference, so the next PR's `expect(handler.predicate).toBe(theNamedPredicate)` identity assertion works uniformly.


## Pull request checklist

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing

#### Release notes

<!-- prettier-ignore -->

### DIFF
--- a/src/confluent/tools/connection-predicates.test.ts
+++ b/src/confluent/tools/connection-predicates.test.ts
@@ -5,6 +5,8 @@ import {
   type PredicateResult,
   ToolDisabledReason,
   allOf,
+  canCreateDirectConnector,
+  flinkWithTelemetry,
   hasCCloudCatalogSupport,
   hasConfluentCloud,
   hasDirectConfluentCloud,
@@ -16,6 +18,7 @@ import {
   hasSchemaRegistry,
   hasTableflow,
   hasTelemetry,
+  kafkaBootstrapOrOAuth,
   widenForOAuth,
 } from "@src/confluent/tools/connection-predicates.js";
 import { describe, expect, it, vi } from "vitest";
@@ -72,6 +75,31 @@ const TABLEFLOW_CONN = conn({
 const CCLOUD_SR_CONN = conn({
   schema_registry: {
     endpoint: "https://psrc-abc.us-east-1.aws.confluent.cloud",
+    auth: { type: "api_key", key: "k", secret: "s" },
+  },
+});
+
+const DIRECT_CCLOUD_KAFKA_AUTH_CONN = conn({
+  confluent_cloud: {
+    endpoint: "https://api.confluent.cloud",
+    auth: { type: "api_key", key: "k", secret: "s" },
+  },
+  kafka: {
+    bootstrap_servers: "broker:9092",
+    auth: { type: "api_key", key: "k", secret: "s" },
+  },
+});
+
+const FLINK_AND_TELEMETRY_CONN = conn({
+  flink: {
+    endpoint: "https://flink.confluent.cloud",
+    auth: { type: "api_key", key: "k", secret: "s" },
+    environment_id: "env-abc",
+    organization_id: "org-123",
+    compute_pool_id: "lfcp-xyz",
+  },
+  telemetry: {
+    endpoint: "https://api.telemetry.confluent.cloud",
     auth: { type: "api_key", key: "k", secret: "s" },
   },
 });
@@ -378,6 +406,82 @@ describe("connection-predicates.ts", () => {
       const widened = widenForOAuth(wrapped);
       widened(OAUTH_CONN);
       expect(wrapped).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("kafkaBootstrapOrOAuth", () => {
+    it("should return ENABLED for a direct connection with kafka.bootstrap_servers", () => {
+      expect(kafkaBootstrapOrOAuth(KAFKA_CONN)).toEqual(ENABLED);
+    });
+
+    it("should report MissingKafkaBootstrap for a direct connection whose kafka block has no bootstrap_servers", () => {
+      expect(kafkaBootstrapOrOAuth(KAFKA_REST_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.MissingKafkaBootstrap),
+      );
+    });
+
+    it("should report MissingKafkaBlock for a direct connection without a kafka block", () => {
+      expect(kafkaBootstrapOrOAuth(SCHEMA_REGISTRY_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.MissingKafkaBlock),
+      );
+    });
+
+    it("should return ENABLED for an OAuth connection (the predicate is widened)", () => {
+      expect(kafkaBootstrapOrOAuth(OAUTH_CONN)).toEqual(ENABLED);
+    });
+  });
+
+  describe("canCreateDirectConnector", () => {
+    it("should return ENABLED when both confluent_cloud and kafka.auth are present on a direct connection", () => {
+      expect(canCreateDirectConnector(DIRECT_CCLOUD_KAFKA_AUTH_CONN)).toEqual(
+        ENABLED,
+      );
+    });
+
+    it("should report MissingConfluentCloudBlock when the first conjunct fails (no confluent_cloud block)", () => {
+      // KAFKA_REST_WITH_AUTH_CONN has kafka.auth but no confluent_cloud,
+      // so hasDirectConfluentCloud short-circuits before hasKafkaAuth runs.
+      expect(canCreateDirectConnector(KAFKA_REST_WITH_AUTH_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.MissingConfluentCloudBlock),
+      );
+    });
+
+    it("should report MissingKafkaBlock when the second conjunct fails (confluent_cloud present but no kafka block)", () => {
+      // CONFLUENT_CLOUD_CONN has confluent_cloud but no kafka block;
+      // hasKafkaAuth's first check (kafka block presence) trips first.
+      expect(canCreateDirectConnector(CONFLUENT_CLOUD_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.MissingKafkaBlock),
+      );
+    });
+
+    it("should report OAuthNotDirectCapable for an OAuth connection", () => {
+      expect(canCreateDirectConnector(OAUTH_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.OAuthNotDirectCapable),
+      );
+    });
+  });
+
+  describe("flinkWithTelemetry", () => {
+    it("should return ENABLED when both flink and telemetry blocks are present on a direct connection", () => {
+      expect(flinkWithTelemetry(FLINK_AND_TELEMETRY_CONN)).toEqual(ENABLED);
+    });
+
+    it("should report MissingFlinkBlock when only the telemetry block is present (first conjunct fails)", () => {
+      expect(flinkWithTelemetry(TELEMETRY_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.MissingFlinkBlock),
+      );
+    });
+
+    it("should report MissingTelemetryBlock when only the flink block is present (second conjunct fails)", () => {
+      expect(flinkWithTelemetry(FLINK_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.MissingTelemetryBlock),
+      );
+    });
+
+    it("should report OAuthNoServiceBlocks for an OAuth connection", () => {
+      expect(flinkWithTelemetry(OAUTH_CONN)).toEqual(
+        disabledFor(ToolDisabledReason.OAuthNoServiceBlocks),
+      );
     });
   });
 });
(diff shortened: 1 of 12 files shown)


src/confluent/tools/connection-predicates.test.ts
```diff
@@ -76,6 +79,31 @@ const CCLOUD_SR_CONN = conn({
   },
 });
 
+const DIRECT_CCLOUD_KAFKA_AUTH_CONN = conn({
```
shouples, line 82: Nit: might be worth renaming `conn()` to `directConnectionConfig()` or similar in a future branch to make it easier to differentiate between it and `OAUTH_CONN` (which could similarly get `oauthConnectionConfig()`)

src/confluent/tools/connection-predicates.ts
```diff
@@ -242,6 +248,37 @@ export function widenForOAuth(
...
 }
 
+/**
+ * The native-Kafka client gate, widened to admit OAuth. Direct connections
+ * still need `kafka.bootstrap_servers`; OAuth connections satisfy it
+ * unconditionally (the broker URL is synthesized from the Auth0 environment).
+ * Use on handlers that have been brought into the OAuth fold.
+ */
+export const kafkaBootstrapOrOAuth: ConnectionPredicate =
+  widenForOAuth(hasKafkaBootstrap);
```
shouples, line 258: Naming here seems... backwards? Almost feels like we should have `hasKafkaBootstrap` as the OAuth-or-direct "widened" variant, and `hasDirectKafkaBootstrap` as the direct-only predicate

shouples (approved):
Looks good and defensive, just some minor suggestions for clearer naming

src/confluent/tools/connection-predicates.ts
```diff
@@ -242,6 +248,37 @@ export function widenForOAuth(
...
+/**
+ * Gate for tools that read telemetry metrics about Flink statements:
+ * requires both the `flink` block (to address a statement) and the
+ * `telemetry` block (to query the metrics API).
+ */
+export const flinkWithTelemetry: ConnectionPredicate = allOf(
+  hasFlink,
+  hasTelemetry,
+);
+
```
jlrobins, line 281: The point of this is to gather them all up into explicitly named *and tested* functions so that those function's unit tests will be authoritative, then removing the need for the referencing handler's tests over `getConnectionIds()`.

Allowing for ad-hoc compositions then does not prove that those ad-hoc-eries are tested as well as their components.

src/confluent/tools/connection-predicates.ts
jlrobins: When oauthery is done, then +10 for @bcscc to do a renaming pass!

src/confluent/tools/connection-predicates.ts
bcscc: Had just understood right after you responded hence deleted comment😅, thanks!

src/confluent/tools/connection-predicates.test.ts
jlrobins: #396

PR #436, opened 2026-05-11
feat: better login pages

### LINKED ISSUE #409: Update to serve custom Confluent branded callback pages after login
See VSCode and IntelliJ examples.

noeldevelops:
## Summary of Changes

- Replace plain string responses with full HTML templates including Confluent logo and styling
- Borrowed the styling & templates from our similar templates in the IntelliJ plugin project
- Closes #409 

Build & Pack to verify template files are included (expect 4 html files including the partial logo & styles html)
``` bash
npm run build
npm pack
tar -tzf confluentinc-mcp-confluent-*.tgz | grep oauth-templates
```
Run the server using the packaged tar created above instead of the published version on npm. Make sure you have configuration for the OAuth flow & use a tool to invoke login and take a look at the new page live. 
```bash
npm install path/to/packaged/tarfile.tgz
npx mcp-confluent -c path/to/config.oauth.yaml 
```
(Depending on your setup and testing env, you may need to run `npm init` and create the `config.oauth.yaml` file as well)

Screenshot 2026-05-11 at 12 31 08 PM

## Pull request checklist

#### Tests

- [X] Added new
- [X] Updated existing
- [ ] Deleted existing

#### Release notes


src/confluent/oauth/callback-pages.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function renderSuccessPage(): string {
+  cachedSuccessPage ??= inlinePartials(readTemplate(SUCCESS_TEMPLATE_FILE));
+  return cachedSuccessPage;
+}
+
+export function renderErrorPage(message: string): string {
+  cachedFailureTemplate ??= inlinePartials(readTemplate(FAILURE_TEMPLATE_FILE));
+  return cachedFailureTemplate.replace("{{error}}", escapeHtml(message));
```
jlrobins: cachedFailureTemplate.replace("{{error}}", escapeHtml(message)) still processes \$\-substitution patterns in the replacement string even when the search value is a literal string. Per ECMA-262, \$\& expands to the matched substring, \$\$ to \$, \$` / \$' to the surrounding template. escapeHtml does not strip \$, and the only attacker-influenceable site — errorParam from pkce-login.ts:127 — is the ?error= query param on the loopback callback. Not exploitable for XSS (the < / > / " / ' escape already covers that), but `?error=foo$&bar` would render the literal text `{{error}}` into the page — a "Wacky" surprise rather than a clean failure message. Fix: pass a replacer function so $ patterns aren't processed.

src/confluent/oauth/pkce-login.ts
```diff
@@ -95,24 +101,34 @@ export async function runPkceLogin(
     rejectFlow = reject;
   });
 
+  const sendHtml = (
```
jlrobins: sendHtml is defined as an arrow inside runPkceLogin but captures nothing from the outer scope. Lifting it to module scope as function sendHtml(...) would shrink the function body and align with the codebase's "function declarations over arrow functions outside of inline callbacks" convention.

jlrobins (commented):
two little tweaks, thanks!

src/confluent/oauth/callback-pages.ts
noeldevelops: ok safer now - `replacement` (2nd param in String.replace) is a function so strings won't get interpolated

src/confluent/oauth/pkce-login.ts
noeldevelops: hoisted!

jlrobins (approved):

jlrobins (approved):

Issue #488, opened 2026-05-19
Consumer-group analysis tooling


jlrobins:
The MCP server can produce to topics, consume from them, and (with #480) read partition watermarks — but it cannot tell a caller what consumer groups exist, who's in one, what they're consuming, or how far behind they are.
Common debugging workflows fall off the cliff:

- "What consumer groups are running on this cluster?"
- "Is group X actually consuming or stuck? Who are its members? What's it assigned to?"
- "What's group X's lag on topic Y right now — the live snapshot, not a metrics time-series?"

The metrics path covers lag as a time-series via the `consumer_lag_offsets` metric (see `docs/workflows/metrics-example-workflows.md`). That answers "plot lag over time for a group I already know about" — not "what groups exist on this cluster?" or "who's in this group right now?" The two views are complementary, not substitutes.

## Approach

Three read-only Kafka tools, all wrapping `KafkaJS.Admin` natively — same client path the rest of the Kafka domain takes (`consume-kafka-messages`, `list-topics`, `create-topics`, and the new `get-partition-offsets` from #480 all go native).

The REST proxy (`/kafka/v3/clusters/{id}/consumer-groups/*`) is available and pre-generated in `src/confluent/openapi-schema.d.ts`, but adopting it for these tools would introduce a second admin surface alongside the native path for no real win. The only thing REST adds is server-side lag arithmetic, which is one `BigInt` subtraction client-side. Native also gives us `listGroups` state filtering that the REST v3 surface doesn't expose.

All three tools share the same wiring decisions:

- Predicate: `kafkaBootstrapOrOAuth` (same as `consume-messages`, `produce-message`, `get-partition-offsets`).
- Category: `ToolCategory.Kafka`.
- Annotation: `READ_ONLY`.
- Standard `cluster_id` / `environment_id` optional args for OAuth connections.
- Response payloads surfaced via `createStructuredResponse` (the `structuredContent` channel #480 introduces).

Issue #435, opened 2026-05-11
Introduce a shared list-pagination helper and adopt across paginated list handlers


shouples:
Follow-up to #129. The bug there (`?cloud=undefined` shipped to CCloud because `cloud` was interpolated into a path key instead of routed through `params.query`) was the latest symptom of a broader fragmentation pattern. Of the 10 list handlers in this repo that touch pagination, only `list-organizations-handler.ts` puts together the full set of "do pagination correctly" decisions:

https://github.com/confluentinc/mcp-confluent/blob/dfa68df44bffe06b38a56de35cc8d617768991cf/src/confluent/tools/handlers/organizations/list-organizations-handler.ts

The other nine sit at different points on the grid:

| Handler                                          | Forwards pageSize/pageToken  | Response Zod | `_meta.nextPageToken`       | try/catch + logger | Notes                                                                                                      |
| ------------------------------------------------ | ---------------------------- | ------------ | --------------------------- | ------------------ | ---------------------------------------------------------------------------------------------------------- |
| `list-organizations-handler.ts`                  | ✅                           | ✅           | ✅                          | ✅                 | Reference impl.                                                                                            |
| `list-flink-statements-handler.ts`               | ✅                           | ❌           | ❌                          | ❌                 | Forwards correctly; no validation, no token surfaced.                                                      |
| `read-flink-statement-handler.ts`                | ✅ (internal loop)           | ❌           | ⚠️                          | ❌                 | Pagination loop is internal; token extraction via string split.                                            |
| `list-tableflow-regions-handler.ts`              | ✅ (post-#129)               | ❌           | ❌                          | ❌                 | Forwards correctly after #129; missing the rest.                                                           |
| `list-tableflow-topics-handler.ts`               | ❌                           | ❌           | ❌                          | ❌                 | **Same shape as #129**: pageSize/pageToken parsed but never forwarded; hardcoded query string in path key. |
| `list-tableflow-catalog-integrations-handler.ts` | ❌                           | ❌           | ❌                          | ❌                 | **Same shape as #129**: same parse-but-drop bug.                                                           |
| `list-clusters-handler.ts`                       | ⚠️ hardcoded `page_size=100` | ✅           | ❌                          | ✅                 | No paging surface at all; hardcoded server-side limit.                                                     |
| `list-environments-handler.ts`                   | ✅                           | ✅           | ⚠️ extracted but unsurfaced | ✅                 | Token pulled but never routed through `_meta`.                                                             |
[...]

PR #492, opened 2026-05-19
List consumer groups tool

### LINKED ISSUE #489: `list-consumer-groups` tool
Smallest of the consumer-group analysis tools.
Wraps `admin.listGroups()` to enumerate consumer groups on a cluster.

State and type filters are exposed as optional args from day one — they save a client-side filter pass and let an LLM caller ask "show me only Stable groups" or "only the empty ones" without a second round trip.

## Tool surface

- Name: `list-consumer-groups`.
- Predicate: `kafkaBootstrapOrOAuth`.
- Category: `ToolCategory.Kafka`.
- Annotation: `READ_ONLY`.

## Input schema

```typescript
z.object({
  matchStates: z
    .array(
      z.enum([
        "Unknown",
        "PreparingRebalance",
        "CompletingRebalance",
        "Stable",
        "Dead",
        "Empty",
      ]),
    )
    .nonempty()
    .optional()
    .describe(
      "Restrict to groups currently in these states. Omit to return groups in any state. " +
        'Common filters: ["Stable"] for live consumers, ["Empty"] for orphaned groups.',
    ),
  matchType: z
    .enum(["Classic", "Consumer"])
    .optional()
    .describe(
      "Restrict to groups of this protocol type. Omit to return both types. " +
        '"Consumer" is the new KIP-848 protocol; "Classic" is the legacy generation-based one.',
    ),
  cluster_id: z
    .string()
    .optional()
    .describe(
      "Confluent Cloud logical Kafka cluster ID (lkc-...). Discover via list-clusters.",
    ),
  environment_id: z
    .string()
    .optional()
    .describe(
      "Confluent Cloud environment ID (env-...) that owns the cluster. Discover via list-environments.",
    ),
});
```

The enum literals must stay in lockstep with the `ConsumerGroupStates` and `ConsumerGroupTypes` enums in `@confluentinc/kafka-javascript/types/rdkafka.d.ts`.
Either hand-list them (as above, easier to read in tool descriptions) or derive them via `z.enum(Object.values(ConsumerGroupStates))` — pick at impl time.

`matchType` is a scalar, not an array — there are only two protocol types in the enum, so a multi-value filter has just one meaningful selection (`[type]`) and one degenerate one (`[both]`, equivalent to omitting the filter).
The underlying `admin.listGroups({matchConsumerGroupTypes})` does take an array; the handler wraps `parsed.matchType` in `[parsed.matchType]` before passing it through.

`.nonempty()` on `matchStates` rejects `[]`, which would otherwise be ambiguous (empty filter = "match nothing" or "match everything"?).

## Response shape

```typescript
type ListConsumerGroupsResponse = {
  groups: Array<{
    groupId: string;
    state: string;
    type: string;
    protocolType: string;
    isSimpleConsumerGroup: boolean;
  }>;
  errors: Array<{ message: string; code?: number }>;
};
```

Text summary: `Found N consumer group(s) on cluster <id>.`
When filters are in play, `Found N consumer group(s) (filtered).` keeps it simple — listing the filter values back is noise.

`errors` carries any per-broker `LibrdKafkaError` entries `listGroups` surfaces as non-fatal partial failures.
Surface them — don't swallow.
A successful call with zero entries is `groups: [], errors: []`, not an error.

- Empty cluster.
  `groups: []`, `errors: []`, no error.
- Multiple groups in mixed states.
  All returned without filtering.
- `matchStates: ["Stable"]` filter narrows correctly.
  Only Stable groups in the response.
- `matchStates: ["Empty"]` filter returns groups with zero members (the orphan case).
- `matchStates: []` (empty array).
  Zod-level rejection — `.nonempty()` catches this with a specific message.
- Invalid state name.
  Zod-level rejection citing the valid values.
- `listGroups` returns partial failure (`errors` array non-empty alongside `groups`).
  Surface the errors verbatim; do not synthesize a tool-level error if at least one group came back.
- `listGroups` returns total failure (`groups: []` and `errors` populated).
  Single tool-level error citing the first broker error message.

## Acceptance criteria

- Handler implemented under `src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts` following the conventions in `.claude/rules/tool-handlers.md`.
- Registered in `tool-name.ts`, `tool-registry.ts`, `index.test.ts`, and `tool-registry.test.ts`'s `EXPECTED_PREDICATES` record.
- Unit tests beside the handler pin every edge case above.
- Integration test under `list-consumer-groups-handler.integration.test.ts` exercises the happy path against a real Confluent Cloud cluster (gated on creds per `.claude/rules/integration-tests.md`).
- CHANGELOG entry.
- README "Kafka tools" section gets a one-line mention.

jlrobins:
## `list-consumer-groups` tool

Wraps `admin.listGroups()` on the native Kafka admin client to enumerate consumer groups on a cluster.

- Predicate `kafkaBootstrapOrOAuth` (same as `consume-messages`, `produce-message`, `list-topics`).
- Category `ToolCategory.Kafka`; annotation `READ_ONLY`.
- Optional `matchStates` (non-empty array of state names) and `matchType` (scalar `Classic` | `Consumer`) save callers a client-side filter pass.
- Standard `cluster_id` / `environment_id` optional args for OAuth connections.

`matchType` is wrapped in a single-element array before being passed through as `matchConsumerGroupTypes`; the underlying `admin.listGroups` accepts an array there but the tool surface deliberately exposes a scalar, since the two-value enum makes a multi-select either redundant or degenerate.

The state and type names surfaced on the response are the TitleCase forms (`Stable`, `Empty`, `Classic`, `Consumer`, …), not the `SCREAMING_SNAKE_CASE` rdkafka enum-key spellings.
librdkafka returns numeric enum values; a forward map translates the Zod-validated input names to numerics on the way in, and a reverse map translates back on the way out.
`Object.values(ConsumerGroupStates)` is intentionally not used — TypeScript's reverse-mapping of numeric enums returns both names and numbers, which Zod would reject.

## `structuredContent` first-mover (alongside `get-partition-offsets`)

This tool surfaces its machine-readable payload via MCP's `structuredContent` channel (per the [2025-11-25 spec](https://modelcontextprotocol.io/specification/2025-11-25/server/tools)).
Adds `BaseToolHandler.createStructuredResponse(text, structured)` next to the existing `createResponse` — additive, no existing handler changes shape.
The text content keeps its human-readable summary role; the broader migration is tracked in #435.

This branch and `feat/480-get-partition-offsets` (still pre-PR) each introduce the same helper independently; whichever lands first wins and the loser drops its copy on rebase.

Wire-format payload:

```typescript
type ListConsumerGroupsResponse = {
  groups: Array<{
    groupId: string;
    state: string;
    type: string;
    protocolType: string;
    isSimpleConsumerGroup: boolean;
  }>;
  errors: Array<{ message: string; code?: number }>;
};
```

Text summary is `Found N consumer group(s).` plain, or `Found N consumer group(s) (filtered).` when any filter is in play.
The original spec carried an `on cluster <id>` clause; dropped because `resolveKafkaClusterArgs` returns `undefined` for direct connections, and the structured payload already carries everything machine-readable.

## Partial- vs total-failure shape

`admin.listGroups` returns `{groups, errors}` where `errors` carries per-broker `LibrdKafkaError` entries surfaced as non-fatal partial failures.
The handler surfaces them verbatim through the `errors` field of the structured payload — partial successes are not collapsed into a tool-level error.

Total failure (`groups: []` and `errors` populated) returns a tool-level `isError: true` response whose text cites the first broker error message.

## Relationships

Closes #489.
Child of #488 (consumer-group analysis tooling).

1. Prompts like "Show me the consumer groups in cluster X" ("What consumer groups are operating in RealWorldData cluster?") while you have some consumers running:
> One consumer group is active on the realworld-data-cluster (lkc-5vmjd8) in the realworld-data-env environment:
>Group ID | State | Type | Protocol
> -- | -- | -- | --
> 3d8e3c56-3c86-4f3a-9cd0-c82e38037805 | Stable | Classic | consumer

<br class="Apple-interchange-newline">

## Pull request checklist

#### Tests

- [x] Added new
- [ ] Updated existing
- [ ] Deleted existing

#### Release notes

- [x] Does anything in this PR need to be mentioned in the user-facing [CHANGELOG](https://github.com/confluentinc/mcp-confluent/blob/main/CHANGELOG.md)?


src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
```diff
@@ -0,0 +1,239 @@
...
+  );
+
+const ENUM_TO_TYPE_NAME: ReadonlyMap<KafkaJS.ConsumerGroupTypes, TypeName> =
+  new Map(
+    (
+      Object.entries(TYPE_NAME_TO_ENUM) as Array<
+        [TypeName, KafkaJS.ConsumerGroupTypes]
+      >
+    ).map(([name, value]) => [value, name]),
+  );
```
bcscc, line 90: should we add a pin test against the enum in the upstream `KafkaJS.ConsumerGroupStates` in case any changes. Something like:
```
  describe("STATE_NAME_TO_ENUM pin", () => {
    it("should cover every numeric value in KafkaJS.ConsumerGroupStates", () => {
      const upstream = Object.values(KafkaJS.ConsumerGroupStates).filter(
        (v): v is KafkaJS.ConsumerGroupStates => typeof v === "number",
      );
      const mapped = new Set(Object.values(STATE_NAME_TO_ENUM));
      const missing = upstream.filter((v) => !mapped.has(v));
      expect(missing).toEqual([]);
    });
  });
```

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
```diff
@@ -0,0 +1,239 @@
...
+
+export const listConsumerGroupsArgs = z.object({
+  matchStates: z
+    .array(z.enum(STATE_NAMES))
+    .nonempty()
+    .optional()
+    .describe(
+      "Restrict to groups currently in these states. Omit to return groups in any state. " +
+        'Common filters: ["Stable"] for live consumers, ["Empty"] for orphaned groups.',
```
bcscc:
```suggestion
      'Filter to groups in any of these states. Omit to return all. ' +
      'Options: "Stable" (live), "Empty" (orphaned), "Dead", ' +
      '"PreparingRebalance", "CompletingRebalance", "Unknown".',
```

Should we list all options for clarity?

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
```diff
@@ -0,0 +1,239 @@
...
+    ),
+  matchType: z
+    .enum(TYPE_NAMES)
+    .optional()
+    .describe(
+      "Restrict to groups of this protocol type. Omit to return both types. " +
+        '"Consumer" is the new KIP-848 protocol; "Classic" is the legacy generation-based one.',
```
bcscc:
```suggestion
      'Filter to groups of this protocol type. Omit to return both. ' +
      'Options: "Consumer", "Classic".',
```

nit/question, is it important to include the additional details for origin of each type?

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  getToolConfig(): ToolConfig {
+    return {
+      name: ToolName.LIST_CONSUMER_GROUPS,
+      description:
+        "Enumerate consumer groups on a Kafka cluster. Read-only — wraps " +
+        "the broker's listGroups admin call. Optional matchStates / matchType " +
+        "filters narrow the result server-side. Response carries one entry " +
+        "per group ({groupId, state, type, protocolType, isSimpleConsumerGroup}) " +
+        "plus an errors array carrying any per-broker partial failures.",
```
bcscc:
```suggestion
        "List consumer groups on a Kafka cluster. Optional filters narrow " +
        "the result server-side. Response carries one entry per group with " +
        "groupId, state, type, protocolType, and isSimpleConsumerGroup, plus " +
        "an errors array of any per-broker partial failures.",
```

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
```diff
@@ -0,0 +1,239 @@
...
+ * input vocabulary; surface it as the literal `"Unknown"` so callers see a
+ * stable string rather than a number.
+ */
+function typeName(value: KafkaJS.ConsumerGroupTypes): string {
+  if (value === KafkaJS.ConsumerGroupTypes.UNKNOWN) return "Unknown";
+  const name = ENUM_TO_TYPE_NAME.get(value);
+  if (name === undefined) {
+    logger.warn(
+      { value },
+      "Wacky -- unrecognized ConsumerGroupTypes numeric value from librdkafka; surfacing as 'Unknown'",
```
bcscc, line 245: I like this error message 😆

README.md
```diff
@@ -321,6 +321,7 @@ get-flink-statement-exceptions: Retrieve the 10 most recent exceptions for a Fli
 get-topic-config: Retrieve configuration details for a specific Kafka topic.
 list-clusters: Get all clusters in the Confluent Cloud environment
 list-connectors: Retrieve a list of "names" of the active connectors. You can then make a read request for a specific connector by name.
+list-consumer-groups: Enumerate consumer groups on a Kafka cluster. Read-only — wraps the broker's listGroups admin call. Optional matchSta...
```
bcscc:
```suggestion
list-consumer-groups: List consumer groups on a Kafka cluster. Read-only — wraps the broker's listGroups admin call. 
```

For consistency

bcscc:
Let's go new tool, thanks @jlrobins, just some minor adjustments & questions

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
jlrobins: New test written.

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
jlrobins: Reworded.

README.md
jlrobins: Reworded.

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
jlrobins: I think the origin detail earns its space: between "Consumer" and "Classic", a caller without prior Kafka context can't tell which is current. Shortened.

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
jlrobins: re-worded.

src/confluent/tools/handlers/kafka/list-consumer-groups-handler.ts
jlrobins: Is longstanding tradition for 'should-never-be-reached' error handling!

bcscc (approved):
Thanks!!

Issue #480, opened 2026-05-18
New `get-partition-offsets` tool to expose per-partition watermarks


jlrobins:
Today the MCP server has no way to answer "what's the current state of this topic's partitions?" without actually consuming. Operators and droids alike reach for that information for every read-side use case that isn't immediate consumption: measuring lag, sanity-checking that a topic is still receiving writes, planning a replay window, sizing a backfill, or just answering "is anything in here?". Today every one of those questions degrades into either "guess a `start: timestamp` and hope" or "consume from `earliest` and count", neither of which actually exposes the underlying numbers.

This issue proposes a small, single-purpose read-only tool that wraps the broker's `listOffsets` admin call. It's the missing primitive that [`consume-messages`](src/confluent/tools/handlers/kafka/consume-kafka-messages-handler.ts) already implicitly relies on but never surfaces.

## What the tool exposes today

The closest neighbors are:

- [`list-topics`](src/confluent/tools/handlers/kafka/list-topics-handler.ts) —
  returns topic names. No partition metadata, no offsets.
- [`get-topic-config`](src/confluent/tools/handlers/kafka/get-topic-config-handler.ts) —
  returns broker-side config (retention, cleanup policy, etc.). Not partition
  state.
- `consume-messages` — does internally compute watermark positions to honor
  `start: "earliest"` / `"latest"` / `{timestamp}`, but doesn't return them.

No tool answers "for partition P of topic T, what are the current low and high watermarks?". That is a one-line `Admin.fetchTopicOffsets(topic)` call on the KafkaJS-compat surface (or `AdminClient.fetchTopicOffsetsByTimestamp` for the timestamp variant), but it isn't exposed.

## Proposed tool surface

New tool `get-partition-offsets` (`ToolName.GET_PARTITION_OFFSETS`), handler at `src/confluent/tools/handlers/kafka/get-partition-offsets-handler.ts`.
Predicate: `kafkaBootstrapOrOAuth` (same as `consume-messages`).
Annotation: `READ_ONLY`. Domain: `ToolDomain.Kafka` (per #419).

Arguments:

| Argument | Type | Notes |
|---|---|---|
| `topicName` | `string` (nonempty) | Single topic per call. |
| `partition` | `number` (optional, >= 0) | If set, restrict to one partition. Otherwise return all partitions of the topic. |
| `cluster_id` | `string` (optional) | Same conventions as `consume-messages`. |
[...]

PR #493, opened 2026-05-19
Implement consume tailmost existing messages.

### LINKED ISSUE #479: Add `tail` mode to `consume-messages` for reading the most-recent already-written message(s)
The seek controls added in #459 cover most of the consumption use cases —
`earliest`, `latest`, absolute offset, by timestamp. But there's one gap
the current `start` tagged union doesn't fill: **"give me the last message
that already exists, without waiting for a new one."**

`start: "latest"` seeks to the partition's high watermark, which is the
offset of the _next_ message to be written. The consumer then blocks until
something is produced. For the interactive "show me the most recent
weather reading" / "what's the last event on this topic?" / "tail the last
5 messages" use cases — the default debug workflow — that means waiting
up to `timeoutMs` for new traffic that may never arrive.

Working around it today is ugly:

- Seek to a recent `{timestamp}`, drain with a big `maxMessages`, keep the
  last. Requires guessing how recent. Costs proportional to how aggressive
  the guess is.
- Seek to `earliest`, consume up to `maxMessages`, hope it's a low-traffic
  topic. Doesn't actually solve the problem on busy topics.
- Compute the watermark ourselves… which we can't, because no tool exposes
  it. (Filed separately as the `get-partition-offsets` companion issue.)

## Proposed surface

Extend the existing `start` tagged union on each `topics[]` entry with
a fourth variant:

```typescript
start?:
  | "earliest"
  | "latest"
  | { offset: string }
  | { timestamp: string | number }
  | { tail: number };     // new — last N already-written messages
```

Concrete Zod fragment for the new variant. The `.describe()` text is the
load-bearing piece here: it has to surface the `partition` requirement
and the positive-integer constraint so LLM callers see both at schema
read time, without a wasted round trip back to the model to learn the
rules:

```typescript
z.object({
  tail: z
    .number()
    .int()
    .positive()
    .describe(
      "Consume the last N already-written messages from this partition. " +
      "REQUIRES `partition` to be set on the same topic entry — there is " +
      "no implicit partition default; consumption is single-partition only. " +
      "N must be a positive integer."
    ),
});
```

The top-level `start` field's umbrella description (which currently
enumerates `earliest` / `latest` / `{offset}` / `{timestamp}`) gets a
trailing clause for `{tail: N}` matching the same style — single
source of truth so a caller scanning either the per-variant or the
umbrella description sees the same rules.

Semantics:

- `{tail: N}` seeks the partition to `max(lowWatermark, highWatermark - N)`
  and consumes forward. The consumer returns whatever is currently sitting
  between that offset and the high watermark — no blocking, no waiting for
  new writes.
- `N` is a positive integer. The Zod fragment uses `.int().positive()`,
  so the schema rejects `0`, negatives, and non-integers up front rather
  than at the handler. No default; explicit value required.
- Requires `partition` to be set on the topic entry, same constraint as
  `{offset}`. The Zod `.describe()` text above states this inline so LLM
  callers don't have to discover the rule by trial-and-error. Multi-partition
  tail semantics are intentionally deferred — see "Out of scope".
- Caps quietly at `highWatermark - lowWatermark` (i.e. asking for `tail: 10`
  on a partition with only 3 retained messages returns 3, not an error).

## Interaction with `maxMessages` / `timeoutMs`

- `maxMessages` continues to bound the total messages returned. A caller
  asking for `tail: 50, maxMessages: 10` gets 10 messages starting at
  `highWatermark - 50`. (Caller's misconfiguration to flag in the schema
  description, not a server-side correction.)
- `timeoutMs` becomes effectively a safety cap rather than the primary
  exit condition: the consumer is reading a bounded already-written
  range, so it should hit `maxMessages` or run out of messages quickly.
  No change to the parameter semantics.

## Implementation notes

The seek dance from #459 mostly transfers. For `{tail: N}`:

1. Subscribe + run + wait for assignment, same as the
   `{offset}` / `{timestamp}` paths.
2. Resolve the high watermark for the requested partition via
   `Admin.fetchTopicOffsets(topic)`.
3. Compute `target = max(low, high - N)` using `BigInt` — Kafka offsets
   are int64 and `N` may be large.
4. `consumer.seek({topic, partition, offset: target.toString()})`, then
   start the message-count loop. Termination is identical to the
   `{offset}` path: hit `maxMessages` or run out of messages (signaled
   by the consumer's iteration ending or hitting the configured
   `timeoutMs`).

Empty partition (`low == high`) just yields zero messages — the seek
target equals the high watermark and there's nothing to consume. Document
this as the expected behavior, not an error.

### Coordination with the `get-partition-offsets` work

The companion `get-partition-offsets` issue proposes a tool that also
wraps `Admin.fetchTopicOffsets`. The two work items may land
concurrently. Whichever lands second must call the watermark-resolution
helper extracted by the first — do **not** ship two independent
`Admin.fetchTopicOffsets` call sites in the codebase. Concretely:

- If `get-partition-offsets` lands first, this issue's handler imports
  the shared helper that landed with it.
- If this issue lands first, the shared helper should land *here*
  (in a sibling module under
  `src/confluent/tools/handlers/kafka/`, not inlined in the consume
  handler) so the `get-partition-offsets` handler can adopt it
  without re-extraction.
- If they land in the same release, the PRs must coordinate on which
  one introduces the helper module and which one adopts it. Reviewers
  on the second PR should reject duplicate watermark plumbing.

The helper itself is small — a thin wrapper around
`Admin.fetchTopicOffsets` returning the normalized
`{partition, low, high}` per-partition shape — but it's the natural
seam between the two handlers and shouldn't be skipped just because
"it's only one call."

- `tail: 1` on a partition with one retained message. Returns that one.
- `tail: 1` on an empty partition. Returns zero, no error.
- `tail: 1000` on a partition with 5 retained messages. Returns 5
  (low-watermark clamp, no error).
- `tail: 1` on a partition where the message at `high - 1` was tombstoned
  / aged out (i.e. `low > high - 1`). Same clamp; returns whatever's at
  or after `low`.
- `tail` with `partition` omitted. Schema-level validation error, same
  message shape as the existing `{offset}` constraint.

## Out of scope

- **Multi-partition tail.** Asking for "the last N messages of this
  topic, across all partitions, freshest-first" requires fetching the
  watermark for every partition, reading from each, then merge-sorting
  by timestamp. Useful, but a bigger surface decision (does freshness
  mean broker timestamp? produce timestamp? deserialized payload
  field?) than belongs in this issue. Requires `partition` for now;
  follow-up if multi-partition demand materializes.
- **Tail-by-time** (`{tail_duration: "5m"}` for "last 5 minutes worth"
  rather than "last N messages"). Expressible today as
  `{timestamp: <now - 5m>}` — separate use case, doesn't need a new
  variant.
- **Streaming/follow mode** (the `tail -f` analog: "give me the last N
  plus anything new for the next M seconds"). The current
  `consume-messages` shape is request/response; introducing follow mode
  is a transport question, not a `start`-variant question.

## Acceptance criteria

- `start` tagged union extended with `{tail: number}` variant, Zod-typed
  as `z.number().int().positive()`.
- The `{tail}` variant's `.describe()` text explicitly states the
  `partition` requirement and the positive-integer constraint, so LLM
  callers see both rules at schema read time.
- The umbrella `start` description (which already enumerates
  `earliest` / `latest` / `{offset}` / `{timestamp}`) is extended to
  mention `{tail: N}` and its `partition` requirement in the same
  style.
- Schema validation rejects `tail` without `partition`, mirroring the
  existing `{offset}` constraint, with a clear error message.
- Server-side resolution calls the shared watermark helper coordinated
  with `get-partition-offsets` (see "Coordination" above) and clamps
  `target = max(low, high - N)`.
- Tests pin: tail on populated partition, tail clamped by low watermark,
  tail on empty partition, tail without partition (schema error), tail
  with N == 0 (schema error), tail with N negative (schema error), tail
  with N non-integer (schema error).
- CHANGELOG entry under the next release.
- README "consume-messages" section updated to document the new
  variant alongside `earliest` / `latest` / `offset` / `timestamp`.

## Related

- #459 — established the `start` tagged union this issue extends.
- Companion issue: `get-partition-offsets` tool, filed separately. The
  two land cleanly together: this issue is the consume-side ergonomic
  sugar, that issue is the standalone metadata primitive. Either can
  ship first; both should exist.
- #438, #475 — Schema Registry decoding policy on `consume-messages`.
  Orthogonal to this work but touches the same handler.
- #476 — `search-topic-messages`. Adjacent in spirit (push the filter
  down into the tool call rather than scan-and-filter in the client),
  but doesn't address the "most-recent already-written message" use
  case this issue targets.

jlrobins:
## `consume-messages` tail mode: `start: {tail: N}`

A fifth arm on the per-topic `start` tagged union, complementing the `"earliest"` / `"latest"` / `{offset}` / `{timestamp}` arms that landed in #483.
Closes the interactive "show me the last N records on this partition" use case that `{timestamp}` and `"latest"` could only awkwardly approximate.

`"latest"` seeks to the partition high watermark and then blocks waiting for new traffic — wrong shape when the caller wants existing records, not future ones.
Working around that today is ugly: guess a recent `{timestamp}` and hope, or drain from `"earliest"` on a low-traffic topic with a generous `maxMessages`.
Tail mode resolves both watermarks server-side and seeks to `max(lowWatermark, highWatermark - N)`, returning immediately with whatever already-written records sit between that offset and the high watermark.

`{tail}` is partition-scoped, same as `{offset}` and for the same reason: "the last N messages of this topic, across all partitions, freshest-first" requires a cross-partition ordering decision (broker timestamp vs producer timestamp vs payload field) that the issue defers as out-of-scope.
The handler rejects `{tail}` without `partition` at pre-flight, mirroring the existing `{offset}` rejection.

## Shared watermark seam: `partition-offsets.ts`

The consume handler had a per-call `createWatermarkCache` wrapping `admin.fetchTopicOffsets`, used from three resolution paths today (explicit-offset bounds check, timestamp's silent-substitution cross-check, earliest-minority low-watermark seek) and now a fourth (tail seek resolution).
Pulled into `src/confluent/tools/handlers/kafka/partition-offsets.ts` so #480's `get-partition-offsets` tool can adopt the same module without re-extraction.

The new module exports:

- `PartitionWatermark` — normalized response shape (`{partition, low, high}`), dropping the upstream `admin.fetchTopicOffsets` `offset` field (the next-to-be-committed group offset, not the partition end-of-log) that neither caller reads.
- `fetchPartitionWatermarks(admin, topic)` — thin wrapper returning `PartitionWatermark[]`.
  #480's standalone path uses this directly; the consume handler reaches it via the cache.
- `WatermarkCache` + `createWatermarkCache(admin)` — the same lazy per-topic memoization the consume handler used inline, now sharable.

Inline `createWatermarkCache` + `WatermarkCache` declarations in `consume-kafka-messages-handler.ts` are gone; the handler imports from the new module.
Colocated `partition-offsets.test.ts` pins both `fetchPartitionWatermarks`'s field-narrowing and `createWatermarkCache`'s one-fetch-per-topic memoization.

## Relationships

- Closes #479.
- Companion: #480 (`get-partition-offsets`).
- Builds on #483, which established the per-topic `start` tagged union this PR extends.

1. Prompts like "`Fetch and decode the most recent 5 messages in WeatherData partition 0. Is in the RealWorldData cluster.`"

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes


src/confluent/tools/handlers/kafka/consume-kafka-messages-handler.test.ts
```diff
@@ -956,6 +1055,141 @@ describe("consume-kafka-messages-handler.ts", () => {
...
+        );
+        admin.fetchTopicOffsets.mockResolvedValue([
+          { partition: 0, low: "10", high: "20", offset: "20" },
+        ]);
+
+        const consumer = await clientManager.getConsumer();
+        consumer.connect.mockResolvedValue(undefined);
+        consumer.subscribe.mockResolvedValue(undefined);
+        consumer.run.mockResolvedValue(undefined);
+        consumer.disconnect.mockResolvedValue(undefined);
```
bcscc, line 1076: nit: Could this be refactored into the before() test block? Seems to be shared across tests?

src/confluent/tools/handlers/kafka/partition-offsets.ts
bcscc, line 1: nit: rename to partition-watermarks to keep consistent with module exports?

bcscc (approved):
Works, thanks! A couple minor nits

PR #495, opened 2026-05-19
Change consume message tool to auto decode schema

### LINKED ISSUE #475: Change consume tool behavior to explicitly disable decoding
Default to always auto decode if SR exists and change the `useSchemaRegistry` into an explicit disable decoding

cc @jlrobins

bcscc:
## Summary of Changes

- Inverts the Schema Registry contract on `consume-messages`: instead of opt-in `useSchemaRegistry: true`, the tool now **auto decodes** when the active connection has SR reachable (a `schema_registry` block on a direct connection, or any OAuth connection), and exposes an opt-out via `disableSchemaRegistry: true` on `valueFormat` / `keyFormat`. 
- Opt-out is per-side — a value-only decode with raw keys still triggers the SR fetch; only when BOTH sides set `disableSchemaRegistry: true` does the handler skip the SR client acquisition entirely.
- SR client connection failures on the auto decode path **fall back to raw bytes** (warn-level log) rather than failing the consume, since the caller didn't explicitly request decoding. Topics without a registered subject and per-message decode errors continue to fall back to raw silently (existing behavior, unchanged).

Fixes #475

Screenshot 2026-05-19 at 2 30 20 PM

1. `npm run inspector -- --config <yaml-with-schema_registry-block>`
2. Call `consume-messages` on an AVRO/JSON/PROTOBUF-encoded topic with `{topics: [{name: "my-topic"}]}` (no `valueFormat` arg) — should return decoded JSON values.
3. Same call with `valueFormat: {disableSchemaRegistry: true}` — should return raw UTF-8 bytes.
4. Same call against a topic with no registered schema — should return raw bytes silently (per-message graceful degradation).
5. Re-run inspector against a config with **no** `schema_registry` block — `consume-messages` should return raw bytes silently and emit no SR-related log lines.

**Before:** Every `consume-messages` call against a fully-configured CCloud cluster needed `valueFormat: {useSchemaRegistry: true}` to get decoded values; the default returned base64-ish raw bytes that an LLM couldn't make sense of without re-calling.

**After:** Bare `{topics: [{name: "my-topic"}]}` does the right thing on an SR-configured connection.

This is a breaking change to the consume tool's argument shape, documented in CHANGELOG.md alongside the other `Unreleased → Changed → consume-messages` breaking shape changes.

#### Tests

- [x] Added
- [x] Updated
- [ ] Deleted existing

#### Release notes

- [x] Documented under \`Unreleased → Changed → consume-messages tool\` in CHANGELOG.md (new bullet on the polarity flip; the prior stale \`{useSchemaRegistry: false}\` reference on the \`valueFormat\`/\`keyFormat\` rename bullet was also corrected).


CHANGELOG.md
```diff
@@ -21,8 +21,9 @@ All notable changes to this MCP server will be documented in this file.
...
+  - **`value` / `key` deserialization options renamed to `valueFormat` / `keyFormat`.** The old names read as data peers to `topics` (especially `value`, which looks like "the value to produce"); the new names describe what the fields do — the format/encoding of the bytes. Both are also now omit-by-default, so the simple consume call no longer needs `value: {}` filler.
   - **`topicNames: string[]` removed in favor of the richer `topics[]` shape.** The single required field is now `topics`, an array of per-topic option objects. The minimal call is `{topics: [{name: "orders"}]}`; the same array carries optional `partition` and `start` controls on the entries that need them.
+  - **Schema Registry decoding now defaults on; opt out per side with `disableSchemaRegistry: true`.** The old `useSchemaRegistry: true` opt-in is replaced with `disableSchemaRegistry: false` (the default) on `valueFormat` / `keyFormat`. When the connection has SR reachable (a `schema_registry` block, or any OAuth connection), messages auto-decode via the registered AVRO / JSON / PROTOBUF schema; topics without a subject and SR client failures fall back to raw silently. Closes #475.
```
jlrobins: Suggest rephrasing:

```suggestion
  - **Schema Registry decoding now defaults on; opt out per side with `disableSchemaRegistry: true`.** The old useSchemaRegistry: true opt-in is gone; SR decoding now happens automatically by default. Opt out per side with `disableSchemaRegistry: true` on valueFormat / keyFormat.
```

jlrobins (commented):
Works great. Just an iota of wordsmitihing.

jlrobins (approved):

PR #498, opened 2026-05-19
Address nits.

jlrobins:
## bcscc nits from #493

Two follow-ups for the post-merge review on #493, both from @bcscc.

## Module rename: `partition-offsets` → `partition-watermarks`

The module exports `PartitionWatermark`, `fetchPartitionWatermarks`, and
`createWatermarkCache`; the file name didn't match. Renamed:

- `src/confluent/tools/handlers/kafka/partition-offsets.ts` → `partition-watermarks.ts`
- The colocated `*.test.ts` and its `describe` label move together.
- `consume-kafka-messages-handler.ts` updates the one import site.

Docstring references to the future `get-partition-offsets` tool (issue #480)
are intentionally left as-is — that's a tool name decision belonging to its
own issue, not file-export naming consistency.

## `describe("handle()")` setup hoisted into `beforeEach`

The 19 success-path tests each opened with the same six lines:

```typescript
const clientManager = getMockedClientManager();
const consumer = await clientManager.getConsumer();
consumer.connect.mockResolvedValue(undefined);
consumer.subscribe.mockResolvedValue(undefined);
consumer.run.mockResolvedValue(undefined);
consumer.disconnect.mockResolvedValue(undefined);
```

Lifted into a single `beforeEach` at the top of `describe("handle()")`, with
suite-scope `let clientManager` / `let consumer` typed via
`ReturnType<typeof getMockedClientManager>` / `ReturnType<typeof getMockedConsumer>`
(no new imports needed). Tests that exercise failure modes
(`consumer.run.mockRejectedValue(...)`, `consumer.disconnect.mockRejectedValue(...)`,
or a custom `consumer.run.mockImplementation(...)` for the
captured-eachMessage case) override the relevant method on the suite-scope
mock and keep working unchanged.

Rebuilt per test rather than initialized once — Vitest's `restoreMocks: true`
only restores `vi.spyOn` originals, not `vi.fn()` call histories or
configured return values, so a `const` at suite scope would leak state
between tests. Matches the pattern documented in `.claude/rules/unit-tests.md`
under "Build per test, not per suite."

Net diff on the test file: +20 / −143.


bcscc (approved):
Thanks!

PR #501, opened 2026-05-20
New "describe consumer group" tool

### LINKED ISSUE #490: `describe-consumer-group` tool
Wraps `admin.describeGroups([groupId])` for a single group and unwraps the result into a flat payload.

Returns the cooked `assignment: TopicPartition[]` per member — not the raw `memberAssignment: Buffer`, which is librdkafka's protocol-level encoded form and is useless to an LLM caller.

## Tool surface

- Name: `describe-consumer-group`.
- Predicate: `kafkaBootstrapOrOAuth`.
- Category: `ToolCategory.Kafka`.
- Annotation: `READ_ONLY`.

## Input schema

```typescript
z.object({
  groupId: z
    .string()
    .min(1)
    .describe(
      "Consumer group ID to describe. Discover existing group IDs via list-consumer-groups.",
    ),
  cluster_id: z
    .string()
    .optional()
    .describe(
      "Confluent Cloud logical Kafka cluster ID (lkc-...). Discover via list-clusters.",
    ),
  environment_id: z
    .string()
    .optional()
    .describe(
      "Confluent Cloud environment ID (env-...) that owns the cluster. Discover via list-environments.",
    ),
});
```

## Response shape

```typescript
type DescribeConsumerGroupResponse = {
  groupId: string;
  state: string;
  type: string;
  protocol: string;
  partitionAssignor: string;
  isSimpleConsumerGroup: boolean;
  coordinator: { id: number; host: string; port: number; rack?: string };
  members: Array<{
    memberId: string;
    clientId: string;
    clientHost: string;
    groupInstanceId?: string;
    assignment: Array<{ topic: string; partition: number }>;
  }>;
};
```

Text summary: `Consumer group "<id>" is <state> with <N> member(s); coordinator <host>:<port>.`

`groupInstanceId` is set iff the member is using static membership — surface only when present, omit the field otherwise.

The raw `memberAssignment` and `memberMetadata` `Buffer` fields the library returns are deliberately dropped from the response.
`assignment` is the cooked equivalent and is what callers want.

- Stable group with multiple members.
  Each member's `assignment` is non-empty and the topic / partition pairs across members are disjoint (no overlap).
- Empty group (state `Empty`, zero members).
  `members: []`; coordinator and state still populated.
- Group ID that doesn't exist on the cluster.
  Clean error: `Consumer group "<id>" not found on this cluster.` — translate the underlying `ERR_GROUP_ID_NOT_FOUND` (or whatever librdkafka surfaces) into a caller-friendly message keyed on the requested name.
- Group with static membership.
  `groupInstanceId` present on members that have one; absent on others.
- Group with the cooperative-sticky assignor.
  `partitionAssignor` reflects it; assignment shape is unchanged.
- `describeGroups` returns a `GroupDescription` whose inner `error` field is set.
  Translate to a tool-level error; don't return a half-populated success payload.
- Group with one member assigned to many partitions across multiple topics.
  `assignment` is flat (topic / partition pairs), not grouped-by-topic — the caller can group if they want.

## Why single-group, not bulk

`describeGroups` natively takes `groupIds: string[]` and could power a bulk variant.
A single-group surface keeps tool arguments predictable and means the LLM never has to disambiguate which group an error in the response refers to.
If real workloads demand bulk, file a follow-up — the cost of fanning out N tool calls in the meantime is low.

## Why not fold lag in

A caller asking "is this group alive and who's in it?" shouldn't pay for an N-topic `fetchTopicOffsets` round-trip just to find out.
Lag is its own composed call (`get-consumer-group-lag`) so the cheap question stays cheap.

## Acceptance criteria

- Handler under `src/confluent/tools/handlers/kafka/describe-consumer-group-handler.ts`.
- Registered in all four wiring points (see `.claude/rules/tool-handlers.md`).
- Unit tests pin every edge case above.
- Integration test against a real Confluent Cloud cluster.
- CHANGELOG entry.
- README mention.

## Relationships

Soft-depends on `list-consumer-groups` for its integration-test setup (need a real group ID to describe) but the handlers are otherwise independent.

jlrobins:
## `describe-consumer-group` tool

Wraps `admin.describeGroups([groupId])` on the native Kafka admin client for a single group and unwraps the result into a flat payload.

- Predicate `kafkaBootstrapOrOAuth` (same as `consume-messages`, `produce-message`, `list-consumer-groups`).
- Category `ToolCategory.Kafka`; annotation `READ_ONLY`.
- Required `groupId` (`.min(1)`); standard optional `cluster_id` / `environment_id` for OAuth connections.

Returns the cooked `assignment` array per member, not the raw `memberAssignment: Buffer` librdkafka exposes at the protocol level (useless to an LLM caller).
`memberMetadata: Buffer` is dropped for the same reason.
`groupInstanceId` is surfaced only when the member is using static membership; otherwise the field is absent.
`coordinator.rack` follows the same conditional shape.

## Workaround for kafkajs `assignment` type/runtime mismatch

Live MCP usage against a real CCloud cluster surfaced that `member.assignment.map is not a function` — the upstream `.d.ts` declares `assignment: TopicPartition[]` (flat array) but the native C++ binding actually returns `assignment: { topicPartitions: TopicPartition[] }` (wrapped object).
Confirmed in `node_modules/@confluentinc/kafka-javascript/src/common.cc` → `FromMemberDescription`, which constructs the wrapped shape verbatim (and also exposes an undocumented optional `targetAssignment: { topicPartitions: ... }` field for KIP-848 rebalance state).

`buildAssignment()` handles both shapes defensively: a flat array passes through (covers the documented shape, in case the binding ever reconciles the lie by fixing the runtime), a `{topicPartitions: [...]}` wrapper unwraps to the same flat result (covers the actual current runtime), and anything else falls through to an empty array with a `"Wacky --"` log line so future surprises stay visible without crashing the tool.
Pinned by three colocated tests: documented flat-array, wrapped-object, and the Wacky fall-through against a `Buffer`.

Follow-up bug report against `confluentinc/confluent-kafka-javascript` filed, https://github.com/confluentinc/confluent-kafka-javascript/issues/487

## Not-found mapping

Three different shapes the broker uses to report an unknown group ID all become the same caller-friendly tool-level error `Consumer group "<id>" not found on this cluster.`:

- `admin.describeGroups` rejects with `LibrdKafkaError.code === ERR_GROUP_ID_NOT_FOUND`.
- The returned `GroupDescription.error` is set with the same code.
- The returned `GroupDescription` is a successful-shaped tombstone: `state: "Dead"`, no members, empty `protocol` / `partitionAssignor`, no `.error` field. Confluent Cloud's brokers were observed taking this path (discovered via integration test against a real cluster); the empty `protocol` and `partitionAssignor` strings are the load-bearing signal that the broker never had a record of the group at all — a previously-alive group that died would still carry its last-used protocol name. Long-tombstoned real groups could in principle match the same fingerprint once the broker drops the retained protocol metadata; accepted as an exotic-edge false-positive in exchange for the friendly UX the issue spec asked for. Pinned by a negative unit test (`Dead` + last-known `protocol: "range"` stays a Dead-but-real success response).

Any other `GroupDescription.error` becomes a tool-level error citing `error.message` verbatim — the handler never returns a half-populated success payload.

## Shared state/type enum mapping

`list-consumer-groups-handler.ts` and `describe-consumer-group-handler.ts` both need to translate the numeric `KafkaJS.ConsumerGroupStates` / `ConsumerGroupTypes` enums to the TitleCase wire strings the response surfaces.
Extracted the maps (forward and reverse) plus the `stateName()` / `typeName()` translators to a new `consumer-group-enums.ts` module so both handlers reference the same source of truth.

The colocated upstream-pin tests (`Object.values(KafkaJS.ConsumerGroupStates)` set-equality assertions, plus the analogous one for `ConsumerGroupTypes` with the deliberate `UNKNOWN` subtraction) move with the maps.
Added two new per-function tests (`stateName()` / `typeName()` mapping every known enum value, plus the wacky-unknown fallthrough) since the translators are now load-bearing for two handlers instead of one.

## Integration test discovers a real group ID

Calling `describe-consumer-group` against the test cluster needs a `groupId` the broker actually knows.
The integration test uses the test-side `withSharedAdminClient` harness to call `admin.listGroups()` once in `beforeAll` and capture the first group's ID; if zero groups exist on the test cluster the describe call mid-test calls `ctx.skip(...)` (a legitimate empty-env outcome, not the kind of broken-env failure that should page someone).
A second `it` block exercises the not-found path against a synthetic group ID so the friendly error message is covered end-to-end regardless of cluster state.

## Text summary

`Consumer group "<id>" is <state> with <N> member(s); coordinator <host>:<port>.`

The structured payload carries everything machine-readable; the text content keeps its human-summary role.

## Relationships

Closes #490.
Child of #488 (consumer-group analysis tooling).
Sibling of #489 (`list-consumer-groups`, merged in #492).

With a real Confluent Cloud Kafka cluster and a known consumer group running, ask: "Describe consumer group `<group-id>` on cluster X".

For example, with a consumer running, I use the prior tool to list the group(s), then use this new tool to examine a single group:

> list the consumer groups in realworld-data-env / realworld-data-cluster

⏺ One consumer group on realworld-data-cluster (lkc-5vmjd8):

| groupId                  | state  | type    | protocolType | simple? |
| ------------------------ | ------ | ------- | ------------ | ------- |
| python-consumer-group-id | Stable | Classic | consumer     | no      |

> Examine that consumer group.

Called confluent (ctrl+o to expand)

⏺ python-consumer-group-id

- State: Stable
- Type: Classic (pre-KIP-848)
- Protocol / assignor: range
- Simple consumer group: no
- Coordinator: broker 1 — b1-pkc-z3k26z.us-east1.gcp.confluent.cloud:9092

Members (1):

| clientId                        | host           | assigned partitions |
| ------------------------------- | -------------- | ------------------- |
| vscode-python-client-5a588608-… | /165.1.202.190 | WeatherData[0]      |

A single VS Code Python client is consuming partition 0 of WeatherData.

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes

- [x] Does anything in this PR need to be mentioned in the user-facing [CHANGELOG](https://github.com/confluentinc/mcp-confluent/blob/main/CHANGELOG.md)?


src/confluent/tools/handlers/kafka/describe-consumer-group-handler.ts
```diff
@@ -0,0 +1,317 @@
...
+
+/**
+ * Wire-format response payload. `Record<string, unknown>` compatibility lets
+ * it thread cleanly through MCP's `structuredContent` channel without a
+ * boundary cast at the call site.
+ *
+ * Deliberately omits the raw `memberAssignment` / `memberMetadata` Buffer
+ * fields the upstream {@link MemberDescription} carries — they're
+ * librdkafka's protocol-level encoded form and useless to an LLM caller.
+ * The cooked {@link TopicPartition}[] on `assignment` is the equivalent
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/confluent/tools/handlers/kafka/describe-consumer-group-handler.integration.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        arguments: { groupId: discoveredGroupId },
+      });
+
+      // The handler emits a `Consumer group "<id>" is <state> ...` prefix on
+      // every success path (Stable / Empty / etc.), and the requested id
+      // appears verbatim in the text — so this proves the tool ran
+      // end-to-end against a real broker AND returned the right group.
+      expect(textContent(result)).toMatch(
+        new RegExp(`^Consumer group "${discoveredGroupId}" is `),
+      );
```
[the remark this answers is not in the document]
jlrobins: Fixed.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this MCP server will be documented in this file.
...
+- **`describe-consumer-group` tool.** Read-only single-group inspector, wrapping the broker's `describeGroups` admin call for one group ID. Returns the group's state, type, protocol, partition assignor, coordinator, and per-member assignment as a cooked `TopicPartition[]` (the raw librdkafka `memberAssignment` / `memberMetadata` Buffers are deliberately dropped — they're protocol-level encoded bytes useless to an LLM). `groupInstanceId` is surfaced only when the member is using static membership; `coordinator.rack` is surfaced only when the broker reports it. Unknown group IDs collapse into the same caller-friendly tool-level error `Consumer group "<id>" not found on this cluster.` across all three broker shapes the handler has observed: a top-level `ERR_GROUP_ID_NOT_FOUND` rejection, a per-group `GroupDescription.error` with that code, and (the path Confluent Cloud actually takes) a successful-shaped tombstone with `state: "Dead"`, no members, and empty `protocol`/`partitionAssignor` strings. Closes #490; second of the consumer-group analysis tools under #488.
```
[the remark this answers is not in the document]
jlrobins: Fixed.

README.md
```diff
@@ -310,6 +310,7 @@ delete-flink-statements: Make a request to delete a statement.
 delete-tag: Delete a tag definition from Confluent Cloud.
 delete-topics: Delete the topic with the given names.
 check-flink-statement-health: Perform an aggregate health check for a Flink SQL statement.
+describe-consumer-group: Describe a single consumer group on a Kafka cluster — wraps the broker's describeGroups admin call for one group ID. Returns state, type, protocol, coordinator, and per-member assignment.
 describe-flink-table: Get full schema details for a Flink table via INFORMATION_SCHEMA.COLUMNS.
```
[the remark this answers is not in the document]
jlrobins: Will revise in a completely separate branch, https://github.com/confluentinc/mcp-confluent/issues/505

src/confluent/tools/handlers/kafka/describe-consumer-group-handler.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  getToolConfig(): ToolConfig {
+    return {
+      name: ToolName.DESCRIBE_CONSUMER_GROUP,
+      description:
+        "Describe a single consumer group on a Kafka cluster — wraps the " +
+        "broker's describeGroups admin call for one group ID. Returns the " +
+        "group's state, type, protocol, partition assignor, coordinator, " +
+        "and per-member assignment as a flat array of {topic, partition} " +
+        "pairs (the raw librdkafka memberAssignment Buffer is dropped).",
```
bcscc:
```suggestion
        "Describe a single consumer group on a Kafka cluster " +
        "Returns the group's state, type, protocol, partition assignor, " +
        "coordinator, and per-member assignment as a flat array " +
        "of {topic, partition} pairs.",
```

Maybe simplified a bit to only describe function?

src/confluent/tools/handlers/kafka/describe-consumer-group-handler.test.ts
```diff
@@ -0,0 +1,725 @@
...
+ *  read tersely; override `code` for other error scenarios. */
+function fakeError(overrides: Partial<LibrdKafkaError> = {}): LibrdKafkaError {
+  return {
+    message: "Broker: Group id not found",
+    code: KafkaJS.ErrorCodes.ERR_GROUP_ID_NOT_FOUND,
+    errno: KafkaJS.ErrorCodes.ERR_GROUP_ID_NOT_FOUND,
+    origin: "kafka",
+    ...overrides,
+  };
+}
```
bcscc, line 74: Similar flavor of these fixtures also exist in`list-consumer-groups-handler.test.ts`, should we try to refactor into a shared kafka test utils file?

bcscc:
Thanks @jlrobins, just have a couple minor comments

src/confluent/tools/handlers/kafka/describe-consumer-group-handler.ts
jlrobins: Done, thanks!

src/confluent/tools/handlers/kafka/describe-consumer-group-handler.test.ts
jlrobins: Good call out. Have factored out `textOf` to a new `call-tool-result.js` and made use of it in the four different files using it so far.

Will do similarly for `fakeError` when the next issue lands and we see what variation it wants, then will know how a central spelling can be made. Right now, two of the implementations disagree some.

bcscc: Wonderful, makes sense thanks!

bcscc (approved):
Thanks!!

PR #504, opened 2026-05-20
chore: add oauth connection to kafka integration tests

### LINKED ISSUE #497: Add integration tests to OAuth supported tools

bcscc:
## Summary of Changes

Adds OAuth coverage to every OAuth-supported kafka handler — 11 from main

- Native-client (`kafkaBootstrapOrOAuth`): `list-topics`, `create-topics`, `delete-topics`, `produce-kafka-message`, `consume-kafka-messages`, `list-consumer-groups`, `get-partition-offsets`, `describe-consumer-group`, `get-consumer-group-lag`
- REST proxy (`kafkaRestWithAuthOrOAuth`): `get-topic-config`, `alter-topic-config`

Follows the dual-mode pattern from #477/#494: each test now has sibling `with a direct connection` / `with an oauth connection` describes under one outer describe.
Direct behavior is preserved verbatim; the OAuth describe is appended as a sibling.

For the long suites (`consume-kafka-messages`, `get-partition-offsets`, `get-consumer-group-lag`) the OAuth describe runs only the basic happy-path assertion.
Per-feature variants (partition pinning, offset/timestamp seeking, mixed-direction, topics filter, error surfaces) stay in the direct describe — those exercise handler logic that doesn't change between auth modes, so duplicating them under OAuth would cost the OAuth flow round trip without buying any new signal.

Closes #497

### `environment_id` under OAuth

`describe-consumer-group` and `get-consumer-group-lag` declare `environment_id` as `.optional()` in their Zod schemas but **require** it at call time under OAuth (the handler resolves the broker from `environment_id` because OAuth connections carry no `kafka` block).
The OAuth describes thread `environment_id: getTestEnvironmentId()` (the YAML-pinned helper from #502) through their `callToolWithOAuthFlow` calls; direct describes are byte-identical to their pre-merge shape.
Future-test note for this rule lives in `.claude/rules/integration-tests.md` under "OAuth-required optional arguments".

### OAuth-only-lane gate

Seeding + verification go through `connectTestAdmin`/`connectTestProducer`/`connectTestConsumer`/`withSharedAdminClient`/`getTestClusterId`, which all build api-key kafka clients from the direct fixture.
To keep an OAuth-only CI run (`INTEGRATION_TEST_CONNECTION_TYPE=oauth`) from crashing in `beforeAll` when direct kafka creds aren't present, each OAuth describe (in the files that seed) gates on `handler.enabledConnectionIds(integrationRuntime({ oauth: false })).length > 0` and skips cleanly via `DIRECT_FIXTURE_REQUIRED_FOR_OAUTH_SEEDING_REASON` (from #502).

`list-topics` and `list-consumer-groups` need no seeding, so they carry no extra gate.

```bash
# both modes (local default)
npm run test:integration -- --tags-filter=@kafka

# OAuth only (requires CONFLUENT_CLOUD_USERNAME + _PASSWORD in .env.integration + `npx playwright-core install chromium`)
INTEGRATION_TEST_CONNECTION_TYPE=oauth npm run test:integration -- --tags-filter=@kafka
```

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes


shouples (approved):
Big batch of test coverage - good stuff!

PR #507, opened 2026-05-20
add npx install command to sign in confirmation screen 

Cerchie:
## Summary of Changes
Adds this copy/pasteable snippet and corresponding telemetry.

Note the use of the timer as a fallback for when user leaves page hanging as well as the templatization of the success page. 
Screenshot 2026-05-21 at 11 20 21 AM

To clicktest, set the mcp to oath mode, trigger the flow by running a tool, and confirm you can copy/paste to install the skills in your terminal. 

## Pull request checklist

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing


assets/oauth-templates/callback-success.html
```diff
@@ -15,5 +15,79 @@ <h1>Authentication Complete</h1>
...
+        var btn = document.getElementById("install-cmd-copy");
+        var cmdEl = document.getElementById("install-cmd");
+        if (!btn || !cmdEl) return;
+        btn.addEventListener("click", function () {
+          var cmd = cmdEl.textContent || "";
+          var done = function (label) {
+            btn.textContent = label;
+            setTimeout(function () {
+              btn.textContent = "Copy";
+            }, 2000);
```
Cerchie, line 72: reverts back to "copy" so user knows they can click again if they like

assets/oauth-templates/callback-success.html
```diff
@@ -15,5 +15,76 @@ <h1>Authentication Complete</h1>
       <p>You have successfully authenticated with Confluent Cloud.</p>
       <p>You may close this page and return to your terminal.</p>
     </div>
+
+    <div class="install-hint">
+      <h2>Next: install the Confluent agent skills</h2>
```
bcscc, line 20: We could add a reference to the agent skills docs for users to be able to get more info? https://docs.confluent.io/cloud/current/ai/ai-tools/agent-skills.html

src/confluent/oauth/pkce-login.ts
```diff
@@ -20,11 +20,30 @@ import {
...
+ * When the tab closes, the stream disconnects and the server tears down.
+ */
+export const SKILLS_HINT_STREAM_PATH = "/skills-hint-stream";
+
+/**
+ * Safety cap on how long the callback server can stay bound after a successful
+ * auth. The stream-disconnect path is the normal closer; this is just the
+ * "user left the tab open and walked away" backstop.
+ */
+export const SUCCESS_PAGE_MAX_LIFETIME_MS = 10 * 60_000;
```
bcscc: Would shortening this to 60–90 seconds be too short? My concern is that leaving the callback page open for any extended period creates a real risk for other stdio sessions and for refreshes on the existing authed session: if a refresh fails non-transiently inside that window and triggers a fresh PKCE, the second attempt errors with `port_in_use: "another mcp-confluent OAuth session may be active"` when in fact it's the same session's leftover listener.

bcscc:
Thanks @Cerchie, great addition, just have 2 questions!

src/confluent/oauth/pkce-login.ts
Cerchie: this makes a lot of sense, can do

src/confluent/tools/handlers/flink/diagnostics/metrics-helper.test.ts
bcscc, line 1: Seems the commit may have picked up some unrelated changes

bcscc (commented):
Thanks for the adjustment, looks good just needs to catch up to main and revert the unrelated metric tests.

src/confluent/tools/handlers/flink/diagnostics/metrics-helper.test.ts
Cerchie: eek, hold on...

Cerchie: ok, caught up for now

bcscc (approved):
Thanks

Issue #564, opened 2026-06-06
Port all of the tools's `handle()` away from the sole-connection guard rail accessors.


jlrobins:
- Port any use of `resolveSoleConnection()` to use `resolveConnection()`, introduced in issue #534, PR #551
- Port any use of `getSoleDirectConnection()` to use ``, , introduced in issue #538, PR #551.

**Completion bar (grep-verifiable).** When this wave lands, the following sole-connection
accessors must have **zero remaining call sites** — they become dead code that will be deleted in a subsequent issue:

- `BaseToolHandler.resolveSoleConnection()` (replaced by §1's `resolveConnection`)
- `ServerRuntime.clientManager` (the singular accessor)
- `MCPServerConfiguration.getSoleConnection()`
- `MCPServerConfiguration.getSoleDirectConnection()`

jlrobins:
All children closed!

Issue #554, opened 2026-06-05
Delete the single-connection scaffolding accessors


jlrobins:
## Context

The single-connection era left behind a family of "sole/singular" accessors that exist
only to let a tool reach _the one_ connection without naming it. As the epic ports handlers
to explicit `connectionId` routing (#534/#538/#539/#541), these become dead code.

#541's completion bar drives them to **zero call sites** but explicitly defers the deletion:
_"they become dead code that will be deleted in a subsequent issue."_ This is that issue.

## Scope: symbols to delete

Once #541 lands, the following must have zero call sites and get removed outright:

- `MCPServerConfiguration.getSoleConnection()` — `src/config/models.ts:209`
- `MCPServerConfiguration.getSoleDirectConnection()` — `src/config/models.ts:230`
- `BaseToolHandler.resolveSoleConnection()` — `src/confluent/tools/base-tools.ts:255`
- `ServerRuntime.clientManager` (the singular getter) — `src/server-runtime.ts:41`

`enforceSingleConnectionOnly()` is **out of scope** — #540 owns its removal.

## Stale prose to clean up in the same sweep

Deleting the symbols isn't enough; the single-connection assumption is narrated in comments
and docstrings that will dangle. Before declaring done, grep the repo and fix each hit:

- `ServerRuntime.clientManager`'s docstring (`src/server-runtime.ts:37-39`) points at #151
  ("Remove … when issue #151 lands") — the wrong epic. It dies with the getter.
- The lifecycle comments narrating the single-connection guard in `server-runtime.ts`
  (around `:47`, `:61-63`, `:74`) reference `enforceSingleConnectionOnly()` / `#151`.
- The two predicate docstrings that describe handlers calling `getSoleDirectConnection()`
  inside `handle()`: `connection-predicates.ts:137` and `:292`.

## Acceptance criteria (grep-verifiable)

1. Zero definitions and zero references for all four symbols:
   ```
   rg "getSoleConnection|getSoleDirectConnection|resolveSoleConnection" src
   rg "\bclientManager\b" src/server-runtime.ts   # singular getter gone; only clientManagers remains
   ```
   returns nothing (test files included).
2. No comment or docstring references any of the four deleted symbols, nor frames the
   single-connection assumption as current behavior.
3. `npm run typecheck` and `npm run test:unit` pass.
4. The server still boots and routes correctly for both single- and multi-connection
   configs (boot coverage established by #540; #554 is dead-code removal only and changes
[...]

Issue #539, opened 2026-06-03
Port the Kafka tool handlers's handle() to use BaseToolHandler.resolveConnection()


jlrobins:
Port the Kafka suite tools away from calling `resolveSoleConnection` to use `resolveConnection()` instead.

Augment test suite proving that these tools do the right thing in multi-connection configurations when invoked against specific connection ids.

Issue #532, opened 2026-06-03
Epic: Multi-connection support


jlrobins:
Let a single server hold multiple connections (the motivating case being Confluent Cloud via OAuth **and** a local Apache Kafka broker side by side) and route each tool call to the right one.

Epic #151 set up most of what we'll need -- YAML configuration document able to describe multiple connections. This epic will remove the guard rails which prevent that document from being used to actually describe multiple connections and the per-tool guard rail accessing the single (configured) connection it was invoked against and will elegantly solve the per-tool cross-cutting issues of:
  * _If this tool can be invoked against more than one connection id in this configuration, define the invocation parameter to do so._
  * _Determine which connection / client manager the tool was actually invoked against._

Two additional tools will be created to let the MCP reason about the connections --- what connections exist and what subset of registered tools are enabled against each, as well as to be able to explain why a given tool was _not_ enabled against a given connection.

Upon completion, the entire codebase will be weaned off of any assumptions about 'at most one connection configured,' and  assuming accessors like:
  * `MCPServerConfiguration.getSoleConnection()`
  * `MCPServerConfiguration.getSoleDirectConnection()`
  * `BaseToolHandler.resolveSoleConnection()`
  * `ServerRuntime.clientManager` (the singular getter)
  
will have zero callers and will be removed!

PR #571, opened 2026-06-08
Port the Connect + Billing tool handlers to routing-aware connection resolution

### LINKED ISSUE #541: Port the Connect + Billing tool handlers' handle() to resolveConnection()
Port the Connect suite (14 handlers) and the Billing tool (`list-billing-costs`) away from `runtime.clientManager` (the singular `ServerRuntime` accessor) / `getSoleConnection()` to `BaseToolHandler.resolveDirectConnection()` (the OAuth-disabled Connect suite) and `resolveConnection()` (the OAuth-enabled Billing tool), reading the caller-supplied `connectionId` off the raw tool arguments and routing to the addressed connection.

Augment the test suite proving these tools route correctly in multi-connection configurations when invoked against specific connection ids — adopt `runtimeWithDecoy` from `tests/factories/runtime.ts` (landed in #539 / PR #566) so existing success cases double as routing tests.

**Handlers (15):** connect (14) + billing (1).

## Relationships

- Child of #564 (port all handlers off the sole-connection accessors), itself a child of #532 (Epic: Multi-connection support).
- Peer of #539 (Kafka suite, the first wave).
- Builds on #551 (`resolveConnection()`).

jlrobins:
## Connect and Billing handlers honor an explicit `connectionId`

The fourteen Connect handlers and the `list-billing-costs` handler previously grabbed `runtime.clientManager` — the singular `ServerRuntime` accessor that throws the moment a runtime holds more than one connection — and the Connect handlers additionally read their connection config via `runtime.config.getSoleDirectConnection()`. Both ignore any caller-supplied `connectionId`. Each now resolves through the seam #551 added, reading `connectionId` off the raw tool arguments and routing to the addressed connection.

The two domains resolve through different doors, and the choice is load-bearing rather than cosmetic:

- **Connect → `resolveDirectConnection`.** Connect tools are OAuth-_disabled_ (they need the Kafka block for env/cluster fallback), so an addressed Connect connection is always `direct`-typed. `resolveDirectConnection` returns a `DirectConnectionConfig`, which is exactly what `resolveConnectEnvAndClusterId(conn, …)` consumes — so the swap preserves the type that `getSoleDirectConnection()` used to hand back, with no narrowing or cast at the call site.
- **Billing → `resolveConnection`.** `list-billing-costs` is OAuth-_enabled_ (`hasConfluentCloudOrOAuth`) and only needs the client manager, never a typed connection. The type-neutral `resolveConnection` is correct here; narrowing to direct would wrongly reject an OAuth connection the tool is happy to serve.

The resolved `{ conn, clientManager }` shape is otherwise unchanged, so each handler's body is a two-line edit: drop the eager `runtime.clientManager`, and replace the sole-connection lookup with a destructure of the routing-aware resolver.

## Routing coverage folded into the existing handle() tests

Routing verification rides on the shared harness rather than a bespoke per-suite test, mirroring #539:

- The eleven Connect suites that drive `handle()` through `assertHandleCase` swap `runtimeWith` for `runtimeWithDecoy` — a one-word change per suite. The decoy connection carries the same service blocks (enabled for the same tool) and is planted first, so a handler that still resolved via `enabledConnectionIds[0]` would land on the decoy; `assertHandleCase` injects the real connection's id and asserts the decoy's client manager went untouched.
- The four suites that call `handle()` directly (`get-connector-error-recommendations`, `get-connector-error-summary`, `get-connector-logs`, and `list-billing-costs`) keep their existing single-connection tests and gain one dedicated routing test each, built on `runtimeWithDecoy`. `get-connector-logs` is the subtle one: its only client-manager touch is auto-resolving the organization id, so that test deliberately omits `organizationId` from both args and config to force the REST call that makes routing observable — otherwise the decoy-untouched assertion would pass vacuously.

These tests are not vacuous: against the un-ported handlers (which call the singular `runtime.clientManager`) the decoy runtime makes every converted case throw "ServerRuntime has multiple client managers," which is precisely the red that flips green once the handler routes by id.

## Remaining sole-accessor call sites

`handle()` in the Connect and Billing domains is now free of the retired accessors. The lone surviving connect-domain `getSoleDirectConnection()` calls are fixture construction in `connect-tool-handler.test.ts` and the integration harness — the same test-side usage left untouched in the config and harness suites, all slated for the repo-wide sweep when #554 deletes the accessors.

## Relationships

- Closes #541.
- Child of #564 (port all handlers off the sole-connection accessors), itself a child of #532 (Epic: Multi-connection support).
- Builds on #551 (`resolveConnection()` / `resolveDirectConnection()`), the seam this PR consumes.
- Peer of #539 (Kafka suite, the first wave), whose `runtimeWithDecoy` harness this PR reuses.

### DIFF
--- a/.claude/rules/unit-tests.md
+++ b/.claude/rules/unit-tests.md
@@ -373,6 +373,22 @@ object. **Do not reach for `vi.mock`** - if a new dependency seems to require it
 - Use `as any` only on partial mock return values (e.g., a mock admin client with only
   `listTopics`), not on the `ClientManager` mock itself; add an eslint-disable comment when needed.
 
+### Routing tests for multi-connection handlers
+
+A handler that resolves its connection via `resolveConnection(runtime, toolArguments)` (or `resolveDirectConnection`) must route to the caller-addressed `connectionId`, not unconditionally to `enabledConnectionIds[0]`.
+`runtimeWithDecoy` (in `tests/factories/runtime.ts`) is a drop-in replacement for `runtimeWith` that proves this: it plants a "decoy" connection — same service blocks, enabled for the same tools, its own auto-minted client manager — inserted _before_ the real one, so a handler that still grabs the first enabled connection lands on the decoy.
+`assertHandleCase` recognizes a decoy-bearing runtime (by the reserved `DECOY_CONNECTION_ID`), auto-injects `connectionId` for the real connection when the caller omitted it, and asserts the decoy's client manager was never touched.
+
+There are two ways to get this coverage, and which one a suite uses depends on how its `handle()` tests are already written:
+
+- **Suites whose `handle()` tests run through `assertHandleCase`** (the `it.each` suites and the per-`it()` `assertHandleCase` suites): swap the suite's `runtimeWith` for `runtimeWithDecoy`.
+  That one-word change turns every existing success case into a routing test — no new test body, and no `connectionId` in the args (the harness injects it).
+  The `it.each` suites need the edit only in the loop's runtime expression; the per-`it()` suites swap each `handle()` runtime.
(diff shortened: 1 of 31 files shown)


flippingbits (approved):
LGTM from reading, thanks!

Issue #568, opened 2026-06-07
Port the Tableflow suite + environments + metrics handlers' handle() to resolveConnection()


jlrobins:
Port the Tableflow suite — `list-tableflow-regions`, 5 `tableflow/catalog` handlers, and 5 `tableflow/topic` handlers, plus the shared `TableflowToolHandler` base — together with the `environments` (2) and `metrics` (2) domains away from `runtime.clientManager` / `getSoleConnection()` to `BaseToolHandler.resolveConnection()`, reading the caller-supplied `connectionId` off the raw tool arguments and routing to the addressed connection.

Augment the test suite proving these tools route correctly in multi-connection configurations when invoked against specific connection ids — adopt `runtimeWithDecoy` from `tests/factories/runtime.ts` (landed in #539 / PR #566) so existing success cases double as routing tests.

**Handlers (15):** tableflow (11) + environments (2) + metrics (2).

## Relationships

- Child of #564 (port all handlers off the sole-connection accessors), itself a child of #532 (Epic: Multi-connection support).
- Peer of #539 (Kafka suite, the first wave).
- Builds on #551 (`resolveConnection()`).

Issue #567, opened 2026-06-07
Port the Flink suite + clusters + organizations handlers' handle() to resolveConnection()


jlrobins:
Port the Flink suite — 5 core statement handlers, 5 `flink/catalog` handlers, and 3 `flink/diagnostics` handlers, plus the shared `FlinkToolHandler` base — together with the single-tool `clusters` and `organizations` domains away from `runtime.clientManager` / `getSoleConnection()` to `BaseToolHandler.resolveConnection()`, reading the caller-supplied `connectionId` off the raw tool arguments and routing to the addressed connection.

Augment the test suite proving these tools route correctly in multi-connection configurations when invoked against specific connection ids — adopt `runtimeWithDecoy` from `tests/factories/runtime.ts` (landed in #539 / PR #566) so existing success cases double as routing tests.

**Handlers (15):** flink (13) + clusters (1) + organizations (1).

## Relationships

- Child of #564 (port all handlers off the sole-connection accessors), itself a child of #532 (Epic: Multi-connection support).
- Peer of #539 (Kafka suite, the first wave).
- Builds on #551 (`resolveConnection()`).

PR #574, opened 2026-06-08
Port the catalog + schema + search handlers to routing-aware connection resolution

### LINKED ISSUE #569: Port the catalog + schema + search handlers' handle() to resolveConnection()
Port the remaining Confluent Cloud catalog and search domains — `catalog` (5 tag handlers), `schema` (2), and `search` (2) — away from `runtime.clientManager` / `resolveSoleConnection()` to `BaseToolHandler.resolveConnection()`, reading the caller-supplied `connectionId` off the raw tool arguments and routing to the addressed connection. This wave closes out the last sole-connection call sites in the handler corpus.

Augment the test suite proving these tools route correctly in multi-connection configurations when invoked against specific connection ids — adopt `runtimeWithDecoy` from `tests/factories/runtime.ts` (landed in #539 / PR #566) so existing success cases double as routing tests.

**Handlers (9):** catalog (5) + schema (2) + search (2).

## Relationships

- Child of #564 (port all handlers off the sole-connection accessors), itself a child of #532 (Epic: Multi-connection support).
- Peer of #539 (Kafka suite, the first wave).
- Builds on #551 (`resolveConnection()`).
- Last handler-porting wave; clears the path for #554 (delete the dead sole-connection accessors).

jlrobins:
# Port the catalog + schema + search handlers to routing-aware connection resolution

This is the final handler-porting wave under #564: with it, every tool's `handle()` resolves its connection through the routing-aware accessors, leaving the sole-connection scaffolding with zero `handle()` call sites for #554 to delete.

## catalog and search handlers honor an explicit `connectionId`

The five catalog (tag) handlers and the two search handlers previously grabbed the eager singular `runtime.clientManager`. Each now resolves through `resolveDirectConnection(runtime, toolArguments)`. They're gated on `hasCCloudCatalogSupport` (which requires a direct `schema_registry` block with api_key auth plus a `confluent_cloud` block) and so are OAuth-disabled — the direct-narrowing resolver is the right one, and none of them need anything off the connection beyond the client manager.

`list-tags` was the one handler whose `handle()` took no `toolArguments` parameter at all (it has no input schema); it gains the standard `(runtime, toolArguments)` signature so it can pass the arguments to the resolver.

## schema handlers honor an explicit `connectionId`

`list-schemas` and `delete-schema` already resolved through `resolveSoleConnection(runtime)`; they swap to `resolveConnection(runtime, toolArguments)`. Schema Registry tools are OAuth-enabled (`hasSchemaRegistryOrOAuth`), so the type-neutral resolver is correct.

## The unit tests catalog and search never had

`schema` already had `handle()` unit tests, so routing rides along by swapping their `runtimeWith` for `runtimeWithDecoy`.

`catalog` had only integration tests and `search` had no tests at all. This wave adds seven `handle()` unit suites — one per tag and search handler — built on `runtimeWithDecoy` so they are routing tests from birth: each asserts the handler routes to the addressed connection and never touches the decoy's client manager, alongside success, API-error, and (where the schema has a required field) `ZodError` cases.

A shared `CATALOG_CONN` fixture joins the factory set: `hasCCloudCatalogSupport` enables a connection only with both a `confluent_cloud` block and a `schema_registry` block carrying api_key auth, so these suites need a connection that actually clears that bar — a plain schema-registry connection would not be a valid route.

## Remaining sole-accessor call sites

`handle()` across catalog, schema, and search is free of the retired accessors. `resolveSoleConnection` still appears on `list-clusters` (handled by the peer #567) and in `base-tools.test.ts` (the method's own test); both clear once their respective waves land and #554 deletes the accessors.

## Relationships

- Closes #569.
- Child of #564 (port all handlers off the sole-connection accessors), itself a child of #532 (Epic: Multi-connection support); this is the last of #564's porting waves.
- Builds on #551 (`resolveConnection()` / `resolveDirectConnection()`).
- Peer of #539 (Kafka), #541 (Connect + Billing), #567 (Flink + clusters + organizations), and #568 (Tableflow + environments + metrics), whose `runtimeWithDecoy` harness and resolver-selection rationale this PR reuses.

### DIFF
--- a/src/confluent/tools/handlers/catalog/add-tags-to-topic.test.ts
+++ b/src/confluent/tools/handlers/catalog/add-tags-to-topic.test.ts
@@ -1,0 +1,76 @@
+import { AddTagToTopicHandler } from "@src/confluent/tools/handlers/catalog/add-tags-to-topic.js";
+import {
+  CATALOG_CONN,
+  DEFAULT_CONNECTION_ID,
+  runtimeWithDecoy,
+} from "@tests/factories/runtime.js";
+import {
+  assertHandleCase,
+  getMockedClientManager,
+} from "@tests/stubs/index.js";
+import { describe, it } from "vitest";
+
+const VALID_ARGS = {
+  tagAssignments: [{ entityName: "lsrc:lkc:orders", typeName: "PII" }],
+};
+
+// Each case runs against runtimeWithDecoy, so assertHandleCase routes to the
+// real connection and asserts the decoy's client manager stays untouched.
+describe("add-tags-to-topic.ts", () => {
+  describe("AddTagToTopicHandler", () => {
+    const handler = new AddTagToTopicHandler();
+
+    describe("handle()", () => {
+      it("should resolve with a success message when the tag is assigned", async () => {
+        const clientManager = getMockedClientManager();
+        clientManager
+          .getConfluentCloudSchemaRegistryRestClient()
+          .POST.mockResolvedValue({
+            data: [{ entityName: "lsrc:lkc:orders" }],
+          });
+        await assertHandleCase({
+          handler,
+          runtime: runtimeWithDecoy(
+            CATALOG_CONN,
+            DEFAULT_CONNECTION_ID,
+            clientManager,
+          ),
+          args: VALID_ARGS,
+          outcome: { resolves: "Successfully assigned tag" },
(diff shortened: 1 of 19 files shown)


bcscc (approved):
LGTM from reading, thanks!

Issue #561, opened 2026-06-05
New per-connection configuration field: `description`


jlrobins:
Augment the yaml configuration structure to allow each connection block to have a `description?: string` field.

Then:
  * Augment `list-connections` output structure to also echo it back (see #536 / #560)
  * Extend example and  fixture yaml configs with some (not all!) carrying description fields
  * Document in the configuration documentation and in the CHANGELOG.

Issue #559, opened 2026-06-05
explain-disabled-tools: regrow output around connections for multi-connection


jlrobins:
### Problem

`explain-disabled-tools` currently emits a single flat list of disabled tools grouped by the
missing config piece — correct under today's single-connection invariant, but lossy once
multiple connections are allowed (epic #532). The handler's own class JSDoc documents the
lossy v1 behavior: a tool enabled on *any* connection is reported as enabled, and a tool
disabled on several connections with different reasons collapses under the first verdict its
iteration produces.

### Proposed v2 (sketched in the handler's class JSDoc)

- Text body becomes one section per connection: a connection header followed by that
  connection's per-gap buckets.
- The structured payload mirrors the same shape; see the `ToolGatingReport` JSDoc in
  `tool-availability.ts` for the target type.

### Relationships

- Child of #532
- Related: #558 (also extends the disabled-reasons report)

Issue #558, opened 2026-06-05
explain-disabled-tools: surface allow/block-list-blocked tools as a disabled reason


jlrobins:
### Problem

`explain-disabled-tools` walks `ToolHandlerRegistry.allHandlers()` and reports each tool's
predicate verdict, but it has no knowledge of the operator's `--allow-tools` / `--block-tools`
filter. A tool the operator explicitly blocked is therefore invisible to (or mis-explained by)
the diagnostic — the symmetric "why is this tool missing?" answer "because the operator's
allow/block list excluded it" is never given.

This is the filter/inventory-tool mismatch flagged while implementing `list-connections` (#536).

### Now-available groundwork

#536 relocated the resolved allow/block-list onto `ServerRuntime` as the single source of truth,
exposed via `runtime.isToolAllowed(name)`. `explain-disabled-tools` can now consult the same
method to fold a filter-blocked bucket into its report.

### Proposed work

- Add a `ToolDisabledReason` (or equivalent grouping key) for "excluded by the operator's
  allow/block list".
- In `buildToolGatingReport` (`tool-availability.ts`), classify a tool as filter-blocked when
  `runtime.isToolAllowed(name)` is false, ahead of the predicate verdict.
- Cover the new bucket in the handler/report tests, pinning the specific reason text.

### Relationships

- Child of #532
- Builds on #536

Issue #550, opened 2026-06-04
Discovery tool: `describe-configured-connection`


jlrobins:
* `describe-configured-connection`: always enabled like #536's `list-connections` (will be renamed to `list-configured-connections` in #562), takes single connectionId as a parameter, enumerates the description, the enabled tools, read-only-ness, AND the disabled tool -> reason why disabled mapping.

Peer tool to #536.

See also issue #365 which overlaps with this issue.

### Implementation note: source the enabled/disabled mapping from `connectionVerdicts()`

Derive the enabled-tools list and the disabled-tool → reason mapping by iterating the
handler catalog and reading each `handler.connectionVerdicts(runtime)`, indexing the
single requested `connectionId` out of the returned map — the same source
`explain-disabled-tools` / `buildToolGatingReport` already consume. Do **not** re-walk
`this.predicate(conn)` independently and do **not** derive the disabled set from
`enabledConnectionIds()` alone (it drops the reasons).

This matters because #570 (per-connection `read_only`) folds its read-only overlay
**into `connectionVerdicts()`** as one additional negative verdict arm
(`ToolDisabledReason.ReadOnlyConnection`). Sourcing from that one method means this
tool honors read-only-ness automatically: a mutating tool on a `read_only` connection
shows up in the disabled mapping with the read-only reason, with no read-only-specific
code here. Group the disabled tools by the opaque `reason` key (never a `switch` over
the enum) so the new arm — and any future verdict arm — surfaces for free.

#570 is slotted to land before this issue; build against the post-#570
`connectionVerdicts()` so the read-only reason is present from the start. #570's
relationship notes also sketch the richer read-only-aware framing this tool can layer
on top (echo the connection-level `read_only` flag; segregate the `ReadOnlyConnection`
bucket; an affirmative "N mutating tools suppressed by read_only" view).

### Relationships

- Child of #532
- Peer of #536 (`list-connections`)
- Builds on #570 (consume the read-only-aware `connectionVerdicts()`)
- See also #365 (overlapping)

PR #578, opened 2026-06-08
Extend per-connection configuration to be able to mark a single connection `read-only`.

### LINKED ISSUE #570: Per-connection `read_only` policy: auto-disable mutating tools on a connection
### Problem / motivation

Operators want to wire one server to several connections with asymmetric trust:
full read/write against a throwaway local-dev broker, but **read-only** against a
staging or prod connection reached over OAuth. Today the only knob with that
flavour is the _server-wide_ `--allow-tools`/`--block-tools` list, which can't say
"writes are fine here but not there" — it's connection-blind.

A per-connection `read_only: true` flag would let an operator grant an agent
generous write access to dev while guaranteeing that the same agent can only ever
_read_ from staging/prod — a meaningful confidence boost for letting an LLM near a
production cluster.

### Why this is not a connection predicate

`ConnectionPredicate` is `(conn: ConnectionConfig) => PredicateResult`: it sees only
the connection, never the tool. Read-only gating is intrinsically cross-cutting —
the verdict depends on **both** the connection's `read_only` flag **and** the tool's
mutation posture. So this must **not** be added to `connection-predicates.ts` or
`EXPECTED_PREDICATES`; it is a separate gating axis layered on top of the predicate.

The tool's mutation posture already exists as `annotations.readOnlyHint`
(`READ_ONLY` ⇒ `true`; `CREATE_UPDATE` and `DESTRUCTIVE` ⇒ `false`). No new per-tool
authoring is required — the existing annotation is the read-only signal.

### Gating axes after this lands

This becomes the third orthogonal gate, peer to the two already in place:

1. Operator allow/block-list — `runtime.isToolAllowed(name)` (tool-only; Pass 1 of
   `getToolHandlersToRegister`).
2. Service-block predicate — `this.predicate(conn)` (connection-only; drives
   `enabledConnectionIds`/`connectionVerdicts`).
3. **Read-only policy × mutation posture** (this issue) — needs both connection and
   tool; layered into the verdict derivation in `BaseToolHandler`.

### Proposed work

**1. Config field.** Add `read_only?: boolean` (default `false`) to both connection
arms — `DirectConnectionConfig` and `OAuthConnectionConfig` — and to
`directConnectionSchema` / `oauthConnectionSchema`. This rides on the same
"per-connection scalar metadata across both union arms" groundwork as the
`description` field (#561); land them in whichever order is convenient, but mirror
that field's placement and test shape.

**2. Verdict overlay in `BaseToolHandler`.** Today `enabledConnectionIds()` and
`connectionVerdicts()` each call `this.predicate(conn)` independently. Make
`connectionVerdicts()` the single source of truth — fold the read-only overlay in
there once — and rewrite `enabledConnectionIds()` to _consume_ it rather than re-walk
the connections. No new helper to name; the overlay lives in exactly one method, and
both stay `@final` (one now derives _through_ the other):

```typescript
connectionVerdicts(runtime: ServerRuntime): Map<string, PredicateResult> {
  return new Map(
    Object.entries(runtime.config.connections).map(([id, conn]) => [
      id,
      this.connectionVerdict(conn),
    ]),
  );
}

enabledConnectionIds(runtime: ServerRuntime): string[] {
  return [...this.connectionVerdicts(runtime)]
    .filter(([, verdict]) => verdict.enabled)
    .map(([id]) => id);
}
```

…where the per-connection composition is inlined in `connectionVerdicts` (or kept as
a tiny private `connectionVerdict(conn)` — the singular of the existing plural, and
"verdict" is already this repo's word for a `PredicateResult`; see the type's own
docstring). The composition is predicate → read-only:

- if `predicate(conn)` is disabled → return that verdict unchanged (a connection
  missing the service block is _not_ "read-only blocked"; service reachability is
  the more fundamental reason and wins);
- else if `conn.read_only === true` **and** this tool's
  `annotations.readOnlyHint !== true` → disabled with the new reason below;
- else → enabled.

Read the mutation posture from `this.getToolConfig().annotations.readOnlyHint`
(cheap — `getToolConfig()` returns a literal).

**3. New `ToolDisabledReason`.** Add a reason named for the condition, e.g.
`ReadOnlyConnection = "connection is marked read_only; tools that mutate state are
disabled"`. (Follow the enum's naming rule: name the symbol after the condition, not
the emitting site.)

**4. No separate call-time check needed.** Because the overlay feeds
`enabledConnectionIds()`, a write tool already (a) drops the read-only connection
from its injected `connectionId` enum, and (b) is fully disabled if its _only_
connection is read-only. `resolveConnection()` already throws
`Wacky -- connection "X" is not an enabled connection for this tool` if an internal
caller forces a disabled id, so registration-layer gating is also the call-time
backstop.

**5. `alwaysEnabled` tools are unaffected.** Connection-agnostic tools (docs,
diagnostics) take no `connectionId` and are all `READ_ONLY` today, so the overlay is
a no-op for them. Pin this in the capstone partition test rather than special-casing
it in the overlay.

### Surfaces that light up for free (verify, don't rebuild)

- `list-connections` (#536): read-only connections show a reduced enabled-tool set.
- `describe-configured-connection` (#550): this is the surface that carries the
  operator-confidence story, so it should do more than echo a flat reason list.
  Build three things from the data already in hand (the connection config + each
  handler's verdict and `annotations.readOnlyHint`):
  1. **Connection-level posture** — surface `read_only: true` as a property of the
     connection itself, not just as a per-tool footnote.
  2. **Segregated disabled bucket** — tools disabled with
     `ToolDisabledReason.ReadOnlyConnection` are exactly the "flip the flag and
     these return" set; keep them distinct from service-block-missing tools (which
     carry `MissingKafkaBlock` etc. — read-only never reaches them).
  3. **Affirmative "writes suppressed" view** — annotate enabled tools with their
     mutation posture and explicitly count the mutating tools the `read_only` flag
     is holding back, e.g. _"14 read tools available; 9 mutating tools suppressed
     by read_only: produce-message, alter-topic-config, …"_. This affirmative
     framing — showing the operator the concrete set of writes that can never fire
     — is the point of the feature for handing an agent a prod connection.
- Startup grouped log (`groupDisabledToolsByReason`) and `explain-disabled-tools`
  (`buildToolGatingReport`): write tools fully disabled by read-only appear under the
  new reason bucket — the exact symmetric extension #558 makes for the allow/block
  axis.
- #559's per-connection report and **cross-connection deltas** section render the
  canonical "staging is read-only; local-dev is full read/write" view with no extra
  work.

### Tests

- `models.test.ts`: `read_only: true` accepted on both arms; defaults to `false`
  when omitted; rejected if non-boolean. Pin the field path.
- `base-tools.test.ts` (verdict derivation): a `READ_ONLY` tool stays enabled on a
  `read_only` connection; a `CREATE_UPDATE`/`DESTRUCTIVE` tool is disabled with
  `ToolDisabledReason.ReadOnlyConnection`; a predicate failure on a `read_only`
  connection still reports the **predicate's** reason, not the read-only one
  (ordering). Use `it.each` over the three annotation kinds.
- `tool-availability` tests: a read-only-blocked write tool lands in the
  `ReadOnlyConnection` bucket of the gating report and the startup log grouping.
- **Do not** add a row to `EXPECTED_PREDICATES` or a case to
  `connection-predicates.test.ts` — this is deliberately not a predicate.

### Docs

- `CONFIGURATION.md`: document `read_only` and its semantics (mutating tools
  auto-disabled per the `readOnlyHint` annotation).
- `config.example.yaml` / `config.oauth.example.yaml`: show `read_only: true` on a
  staging/prod-flavoured connection (pairs naturally with the OAuth example).
- CHANGELOG: one-line headline bullet.

### Relationships

- Child of #532
- Builds on #561 (per-connection metadata field groundwork)
- Related: #558 and #559 (both extend the disabled-tools report around new axes;
  this feature is a first-class producer of a per-connection disabled reason)
- Related: #550 — `describe-configured-connection` is the primary beneficiary; it
  should surface the connection-level `read_only` flag, segregate the
  `ReadOnlyConnection` disabled bucket from service-block-missing tools, and render
  the affirmative "mutating tools suppressed by read_only" view (see Surfaces above).

jlrobins:
# Per-connection `read_only` policy

## Read-only connection policy

- Operators can now mark a connection `read_only: true` to guarantee an agent can only ever read from it — the natural posture for handing an LLM a staging or prod connection while keeping full read/write on a throwaway dev broker. Server-wide `--allow-tools`/`--block-tools` couldn't express this: it's connection-blind, whereas read-only gating needs both the connection's flag and the tool's mutation posture.
- This is the third orthogonal gating axis, peer to the operator allow/block-list and the service-block predicate. It is deliberately **not** a `ConnectionPredicate`: a predicate sees only the connection, never the tool, and read-only is intrinsically cross-cutting. No row was added to `EXPECTED_PREDICATES`.
- The tool's mutation posture is read straight off the existing `annotations.readOnlyHint` (`READ_ONLY` ⇒ `true`; `CREATE_UPDATE`/`DESTRUCTIVE` ⇒ `false`). No new per-tool authoring.

## Config field

- `read_only?: boolean` rides on both connection arms (`DirectConnectionConfig` / `OAuthConnectionConfig`) and both Zod schemas, mirroring the `description` groundwork from #561. The schema applies `.default(false)`; the interface field is optional so the legacy env synthesizer and the test fixtures need no change.
- `models.test.ts` pins the field path: accepted on both arms, defaults to `false` when omitted, rejected (with `invalid_type`) for a non-boolean.

## Verdict overlay in `BaseToolHandler`

- `connectionVerdicts()` is now the single source of truth for per-connection enablement; `enabledConnectionIds()` derives through it rather than re-walking the connections. Both stay `@final`. A new private `connectionVerdict(conn)` composes the two layers.
- The composition is predicate → read-only: a connection the predicate already disables keeps the predicate's (more fundamental) reason; an enabled connection that is `read_only` disables mutating tools with the new `ToolDisabledReason.ReadOnlyConnection`.
- Because the overlay feeds `enabledConnectionIds()`, write tools drop the read-only connection from their injected `connectionId` enum and fully disable when read-only is their only connection — no separate call-time check needed; `resolveConnection()` remains the backstop.
- `base-tools.test.ts` drives an `it.each` over the three annotation kinds against a `read_only` connection (read-only stays enabled; the two mutating kinds disable with `ReadOnlyConnection`) plus an ordering case proving a predicate failure on a read-only connection still reports the predicate's reason.

## Surfaces that light up for free (verified, not rebuilt)

- `list-connections` shows a reduced enabled-tool set for a read-only connection — pinned in `list-connections-handler.test.ts`.
- The startup grouped log (`groupDisabledToolsByReason`) and `explain-disabled-tools` report (`buildToolGatingReport`) surface read-only-disabled write tools under the new `ReadOnlyConnection` bucket — pinned in `tool-availability.test.ts`.
- Connection-agnostic (`alwaysEnabled`) tools are a no-op for the overlay because every one is `READ_ONLY`; rather than special-case the overlay, this invariant is pinned in `tool-registry.test.ts` (a future mutating `alwaysEnabled` tool fails it).

## Docs

- `CONFIGURATION.md`, `config.example.yaml`, and `config.oauth.example.yaml` document the flag (the OAuth example carries the prod-flavoured `read_only: true` illustration); one-line CHANGELOG headline.

## Relationships

- Closes #570
- Child of #532 (multi-connection support epic)
- Builds on #561 (per-connection scalar-metadata groundwork — the `description` field)
- Related: #558 / #559 (both regrow the disabled-tools report around new axes; this feature is a first-class producer of a per-connection disabled reason) and #550 (`describe-configured-connection`, the primary future beneficiary — out of scope here)


Cerchie (approved):
LGTM. Tested with read_only set and not set, then with oauth v env secrets. All as expected. Just needs update w main. 

_explain_disabled_tool_reason_ gives the expected response: 
Image


no _read_only_ set:
Image

_read_only_ set in config: 

Image

jlrobins:
Easy conflict resolution, thanks @Cerchie !

Cerchie (approved):

PR #580, opened 2026-06-10
 Make the four Kafka direct-call suites uniform with the rest of the corpus

### LINKED ISSUE #565: Make the four Kafka direct-call suites uniform with the rest of the corpus
Almost every handler `handle()` suite runs its cases through `assertHandleCase` against a
`runtimeWithDecoy` runtime. Four Kafka suites are the outliers — `list-consumer-groups`,
`describe-consumer-group`, `get-consumer-group-lag`, `get-partition-offsets` — which drive
`handle()` directly and bolt on a single routing test each. Bring them into the same shape
as everyone else.

As a side benefit, routing coverage improves: today each suite proves routing in one test
while its other cases assert nothing about which connection was hit. Once every case runs
through the harness, every case also asserts the decoy stayed untouched — so a handler that
routed to the wrong connection on any path would fail, not just on the one bespoke test.

## Tasks

- Make `assertHandleCase` return the resolved `CallToolResult` (`undefined` if the handler
  threw) so a case can assert exact `structuredContent` or `textOf` after the harness call.
- In each of the four suites, run the behaviour cases through `assertHandleCase` against a
  `runtimeWithDecoy` runtime, and delete the dedicated routing test.
- Leave as direct `handle()` calls the cases the harness can't express:
  - cases pinning `code` / `errno` / `origin` on a thrown error via `.rejects.toMatchObject`
    (the `{ throws }` matcher only sees the message);
  - Zod-boundary cases pinning `issues[].path` / `issues[].code` (the matcher collapses any
    `ZodError` to `"ZodError"`).
- Update the routing section of `.claude/rules/unit-tests.md`.

jlrobins:
## Uniform handler-test shape across the Kafka surface

Almost every handler `handle()` suite runs its cases through `assertHandleCase` against a `runtimeWithDecoy` runtime. Four Kafka suites were the holdouts — `list-consumer-groups`, `describe-consumer-group`, `get-consumer-group-lag`, `get-partition-offsets` — driving `handle()` directly in their behaviour tests and bolting on a single bespoke routing test each. This PR routes their behaviour cases through the harness too, so the four now read like every other suite and the four bespoke routing tests are deleted. A reader moving between the Kafka, Flink, and Tableflow suites sees one idiom instead of two.

A few cases stay as direct `handle()` calls, because the harness's outcome matchers can't express what they pin — and these are the same principled exceptions any suite would carry:

- `get-consumer-group-lag` keeps three cases that assert `code` / `errno` / `origin` on a thrown librdkafka error via `.rejects.toMatchObject`; the harness's `{ throws }` matcher only sees the message.
- `get-partition-offsets` keeps three Zod-boundary cases that pin `issues[].path` / `issues[].code`; `classifyThrown` collapses every `ZodError` to `"ZodError"`, and these throw before connection resolution anyway.

## Routing coverage improves as a side benefit

Each suite previously proved routing in exactly one test while its other behaviour cases asserted nothing about which connection was hit. Now every routed case asserts the decoy connection's client manager stayed untouched, so a handler that routed to the wrong connection on any path fails — not just on the one bespoke test. The tripwire proves it isn't vacuous: reverting `ListConsumerGroupsHandler` to grab the first enabled connection turns 9 converted cases red on the decoy-untouched assertion.

## `assertHandleCase` hands back the result

`assertHandleCase` now returns the resolved `CallToolResult` (`undefined` if the handler threw), so the converted cases that pin an exact `structuredContent` shape or an exact/regex `textOf` capture the return and assert after the harness call — the `resolves` substring alone is too coarse for these suites. Existing callers that ignore the return are unaffected.

The routing section of `.claude/rules/unit-tests.md` is updated to match: the harness is the one way to get routing coverage, the consumer-group/offsets suites are no longer the direct-call exemplar, and the two case kinds that stay direct are spelled out.

## Relationships

Closes #565. Member of under epic #532.

### DIFF
--- a/.claude/rules/unit-tests.md
+++ b/.claude/rules/unit-tests.md
@@ -378,16 +378,22 @@ object. **Do not reach for `vi.mock`** - if a new dependency seems to require it
 A handler that resolves its connection via `resolveConnection(runtime, toolArguments)` (or `resolveDirectConnection`) must route to the caller-addressed `connectionId`, not unconditionally to `enabledConnectionIds[0]`.
 `runtimeWithDecoy` (in `tests/factories/runtime.ts`) is a drop-in replacement for `runtimeWith` that proves this: it plants a "decoy" connection — same service blocks, enabled for the same tools, its own auto-minted client manager — inserted _before_ the real one, so a handler that still grabs the first enabled connection lands on the decoy.
 `assertHandleCase` recognizes a decoy-bearing runtime (by the reserved `DECOY_CONNECTION_ID`), auto-injects `connectionId` for the real connection when the caller omitted it, and asserts the decoy's client manager was never touched.
+It returns the resolved `CallToolResult` (`undefined` if the handler threw), so a case that pins an exact `structuredContent` shape or an exact/regex `textOf` — beyond the `resolves` substring — can capture the return and assert after the harness call.
 
-There are two ways to get this coverage, and which one a suite uses depends on how its `handle()` tests are already written:
+To get routing coverage, run a suite's `handle()` cases through `assertHandleCase` with a `runtimeWithDecoy` runtime.
+That turns every case into a routing test — no `connectionId` in the args (the harness injects it), and no bespoke routing test body.
+The `it.each` suites need the edit only in the loop's runtime expression; the per-`it()` suites build each case's runtime with `runtimeWithDecoy`.
+The consumer-group and offsets suites (`list-consumer-groups`, `describe-consumer-group`, `get-consumer-group-lag`, `get-partition-offsets`) follow this shape — their behaviour cases all run through the harness, capturing the returned result for the shape/`textOf` assertions the substring check can't express.
 
-- **Suites whose `handle()` tests run through `assertHandleCase`** (the `it.each` suites and the per-`it()` `assertHandleCase` suites): swap the suite's `runtimeWith` for `runtimeWithDecoy`.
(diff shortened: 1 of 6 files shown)


.claude/rules/unit-tests.md
```diff
@@ -378,16 +378,22 @@ object. **Do not reach for `vi.mock`** - if a new dependency seems to require it
...
 `assertHandleCase` recognizes a decoy-bearing runtime (by the reserved `DECOY_CONNECTION_ID`), auto-injects `connectionId` for the real connection when the caller omitted it, and asserts the decoy's client manager was never touched.
+It returns the resolved `CallToolResult` (`undefined` if the handler threw), so a case that pins an exact `structuredContent` shape or an exact/regex `textOf` — beyond the `resolves` substring — can capture the return and assert after the harness call.
 
-There are two ways to get this coverage, and which one a suite uses depends on how its `handle()` tests are already written:
+To get routing coverage, run a suite's `handle()` cases through `assertHandleCase` with a `runtimeWithDecoy` runtime.
+That turns every case into a routing test — no `connectionId` in the args (the harness injects it), and no bespoke routing test body.
+The `it.each` suites need the edit only in the loop's runtime expression; the per-`it()` suites build each case's runtime with `runtimeWithDecoy`.
```
Cerchie, line 385: 💥

flippingbits (approved):

PR #582, opened 2026-06-11
build: migrate package manager from npm to pnpm

cchristous:
## Summary

Migrate this repo's **package manager from npm to pnpm**. This is a package-manager swap only — no runtime, bundler, or test-runner change, and the published package is unaffected.

Motivation: pnpm gives faster, content-addressable installs (notably in CI), strict dependency resolution that rejects phantom/undeclared dependencies, and disk-efficient hardlinked stores.

## What changed

- **`package.json`**: pin `packageManager: pnpm@10.33.4`; move the npm top-level `overrides` to `pnpm.overrides`; add `pnpm.onlyBuiltDependencies` for the native/build-script deps (`@confluentinc/kafka-javascript`, `protobufjs`, `fsevents`) since pnpm v10 does not run dependency build scripts by default; **declare `zod` as a direct dependency** (see Findings); drop `engines.pnpm` so the dev-only pnpm constraint doesn't ship in the published manifest (`packageManager` remains the source of truth and is stripped from the tarball).
- **Lockfile**: replace `package-lock.json` with `pnpm-lock.yaml`. `.gitignore` previously blanket-ignored root `*.yaml`, so the lockfile is explicitly un-ignored; npm/yarn lockfiles are now ignored so they can't be reintroduced.
- **Semaphore pipelines** (`semaphore.yml`, `integration.yml`, `release-version.yml`, `npm-audit.yml`): activate the pinned pnpm via `corepack enable`, replace `npm ci` with `pnpm install --frozen-lockfile`, and cache the pnpm store keyed on the lockfile checksum (stored on `epilogue.always` for warm-cache resilience). `publish-npm.yml` intentionally stays on `npm publish` — it only uploads a prebuilt tarball.
- **`Dockerfile`**: builder + multi-stage migrated to pnpm. pnpm uses relative symlinks with the `.pnpm` store nested under `node_modules`, so the pruned `node_modules` copies cleanly into the production stage; `npm prune --omit=dev` → `pnpm prune --prod`.
- **Husky hooks, Makefile, and docs** (README, CONTRIBUTING, CLAUDE.md, `.claude/rules`, copilot-instructions, PR template) updated to pnpm. The npm-registry badge and the `npx @confluentinc/mcp-confluent` end-user invocations are unchanged — the package still ships to npm.

## Findings (why these changes are correct)

- **`zod` was a phantom dependency**: it is imported in ~69 source files but was never declared in `dependencies` (only force-resolved via `overrides`). npm's flat `node_modules` hoisted it so imports happened to resolve; pnpm's strict layout correctly refused, surfacing the latent bug. Now declared at `^4.0`.
- **Native build allowlist is complete**: a scan of the resolved tree for install/postinstall scripts yields exactly the three allowlisted packages.

## Verification

- `pnpm install --frozen-lockfile`, `build`, `lint`, `typecheck`, `test:harness` (18), and `test:unit` (1996) all pass locally; the native `@confluentinc/kafka-javascript` binary builds and the CLI boots.
- `pnpm pack` produces an identically-named tarball (`confluentinc-mcp-confluent-<version>.tgz`) with the same `files` set, so the publish path and downstream consumers are unaffected.
- All four migrated Semaphore YAML files validated (parse + unique-key check).

## CI scope on this PR

This PR touches no `src/**` files, so the `change_in`-gated integration and smoke blocks are skipped; only the **Build & Test** block runs (which needs no credentials).

> Draft while before/after CI install-timing is collected for a pnpm-vs-npm comparison.


cchristous:
Superseded by #583, which ports the same commit to a branch in the confluentinc fork (this PR was opened from a personal fork by mistake).

PR #591, opened 2026-06-12
Remove the single-connection guard and allow multiple connections

### LINKED ISSUE #540: Remove the single-connection guard and allow multiple connections
Remove `enforceSingleConnectionOnly()` so a config with more than one connection parses and
boots. The runtime is already multi-connection-ready: `ServerRuntime.fromConfig()` builds a
client manager per connection, and every handler resolves its target via `resolveConnection`
/ `resolveDirectConnection`. This guard is the last thing blocking a multi-connection config.

## Where it lives

- `enforceSingleConnectionOnly()` is defined in `src/config/models.ts` and invoked from the
  `MCPServerConfiguration` validation there — **not** in `main()` / `src/index.ts`. It rejects
  any config carrying more than one connection at parse time.

## Tasks

- Delete `enforceSingleConnectionOnly()` and its invocation in `src/config/models.ts`.
- Add minimal tests proving the server boots when configured for multiple connections — both
  direct + direct and direct + oauth.
- Fix the comments and docstrings that narrate the guard as current behaviour; they go stale
  the moment it's removed:
  - `src/server-runtime.ts` (around the multi-manager guards and the per-connection
    construction loop)
  - `src/confluent/tools/tool-availability.ts`
  - `src/confluent/tools/handlers/diagnostics/explain-disabled-tools-handler.ts`

## Out of scope

- The singular `ServerRuntime.clientManager` getter in `src/server-runtime.ts` is the runtime
  mirror of this guard. It already has zero callers; deleting it belongs to #554, not here.
- Keep the multiple-OAuth-connections rejection in `ServerRuntime.fromConfig()` — that's a
  separate, permanent constraint (only one OAuth connection is supported), not the
  single-connection guard.

## Result

Clear to start clicktesting with multiple connections!

jlrobins:
# Remove the single-connection guard and allow multiple connections

## Guard removal

- Deleted `enforceSingleConnectionOnly()` from `src/config/models.ts` and its `.refine()` on the
  `connections` record. The guard is gone in both directions: a config may now define more than one
  connection, and it may define zero. There is no replacement `>= 1` refine — an empty `connections: {}`
  is a valid, bootable config (see "Connection-independent enablement" below for why that's useful).

## Per-connection client-manager construction

- `ServerRuntime.fromConfig()` previously handed _every_ connection an `OAuthClientManager` whenever
  _any_ connection was OAuth. In a mixed config (the epic's motivating "Confluent Cloud via OAuth +
  local Apache Kafka broker" case) the direct connection silently got an OAuth manager pointed at
  CCloud and could never reach its broker. The construction loop now branches per connection on
  `conn.type`: direct connections get `constructDirectClientManager(conn)`, the lone OAuth connection
  gets an `OAuthClientManager` backed by the shared `OAuthHolder`.
- Boot tests cover `direct + direct` and `direct + oauth`, asserting each connection's concrete
  manager type — so a regression back to the all-OAuth loop fails loudly rather than booting a
  broken mixed config.
- Server teardown disconnected only the _sole_ client manager via the singular
  `ServerRuntime.clientManager` getter, which throws on a multi-connection config — so a
  multi-connection server crashed at shutdown. Added `ServerRuntime.disconnectAll()` (disconnects
  every per-connection manager) and routed `performCleanup` through it. The singular getter now has
  no production callers; its deletion stays with #554.
- `disconnectAll()` uses `Promise.allSettled`, not `Promise.all`: one manager's `disconnect()`
  rejection must not abandon its siblings mid-shutdown. It waits for every disconnect, then throws an
  `AggregateError` carrying all failures so none is silently dropped. Tests assert both the
  single-failure and multi-failure aggregation shapes.
- `performCleanup` is now a module-scope, exported function that runs every teardown step inside a
  `try/catch/finally` and exits in the `finally`. A rejection from any step (most plausibly a manager
  failing to disconnect) is logged but never strands the process — the exit always fires. `exit` is
  injected (defaulting to `process.exit`) so the determinism guarantee is unit-testable without
  stubbing the global; tests cover both the clean path and a rejecting step.

## Connection-independent enablement

- Four tools carry the `alwaysEnabled` predicate and work without any connection: `search-product-docs`,
  `get-product-doc-page`, `list-configured-connections`, and `explain-disabled-tools`. Enablement was
  derived per-connection, so a zero-connection config left even these with no enabled connection IDs —
  the server registered nothing and refused to boot, and `explain-disabled-tools` would itself throw on
  an empty verdict map.
- `BaseToolHandler` now exposes `isConnectionIndependent` (≡ the predicate is `alwaysEnabled`, the only
  unconditional predicate). Registration enables a tool when it has at least one enabled connection
  **or** it is connection-independent, so an empty config boots with exactly those four tools available.
- Added `ToolDisabledReason.NoConnectionsConfigured`. On a zero-connection config, `explain-disabled-tools`
  buckets every connection-dependent tool under it instead of crashing. The report builder now classifies
  each handler in one place (`classifyHandler`): connection-independent → enabled, zero connections →
  `NoConnectionsConfigured`, otherwise the per-connection verdicts decide (`classifyVerdicts`, tightened to
  a non-empty-map contract). The old `Wacky --` empty-verdict-map guard is deleted — its justification
  ("`enforceSingleConnectionOnly` rejected zero connections at bootstrap") no longer exists, and the
  empty-map case is now a handled, legitimate state rather than an invariant violation.

## OAuth login deferred to first OAuth-routed call

- In a mixed `local + ccloud-oauth` config the browser OAuth bounce fired on the _first tool call
  of any kind_, including a purely-local Kafka call — the gate in `mcp/server.ts` keyed on "an
  OAuth connection exists in the config AND the tool isn't connection-independent," never on which
  connection the call actually targets.
- New `BaseToolHandler.resolvedTargetConnectionId(runtime, args)` reports the connection a call
  routes to. It delegates to `resolveConnection` so routing stays decided in one place rather than
  duplicated in the gate; it returns `undefined` for connection-independent tools and for
  unresolvable/ambiguous calls (which `handle()` then reports the routing error for). The gate now
  calls `ensureLoggedIn()` only when that target connection is OAuth-typed — so a local-routed call
  never launches the browser, and the bounce happens on the first call that genuinely targets the
  OAuth connection.
- Tests: the OAuth-login gate suite now drives real routing (local-routed → no login; OAuth-routed,
  both implicit-sole and explicit-`connectionId` → login), plus focused `resolvedTargetConnectionId`
  unit tests covering the connection-independent, sole-enabled, explicit, ambiguous, and
  unknown-id branches.

## Stale single-connection claims swept

- After fixing the teardown crash (which the issue had wrongly described as having "zero callers"),
  swept the codebase for other code or prose assuming exactly one connection/manager. The logic is
  clean: every connection/manager walk iterates the full record (`config-telemetry.ts` already emits
  an array of connection types; `server-runtime.ts`, `index.ts`, `base-tools.ts`, and
  `list-configured-connections-handler.ts` all use `Object.entries/values`), and the `getSole*` /
  `resolveSoleConnection` accessors have no production callers (they wait on #541/#554).
- One prose finding, fixed: two predicate docstrings in `connection-predicates.ts` (`hasConfluentCloud`,
  `canCreateDirectConnector`) still said handlers "call `getSoleDirectConnection()` inside `handle()`."
  #564 ported every connect handler to `resolveDirectConnection()`, so the docstrings now name that.

## Relationships

- Closes #540
- Child of #532 (Epic: Multi-connection support)

### DIFF
--- a/src/confluent/tools/base-tools.ts
+++ b/src/confluent/tools/base-tools.ts
@@ -123,6 +123,12 @@ export interface ToolHandler {
    * `describe-tool-availability` diagnostic tool.
    */
   connectionVerdicts(runtime: ServerRuntime): Map<string, PredicateResult>;
+
+  /**
+   * Whether this tool is enabled irrespective of the configured connections.
+   * See {@linkcode BaseToolHandler.isConnectionIndependent}.
+   */
+  readonly isConnectionIndependent: boolean;
 }
 
 export interface ToolConfig {
@@ -178,10 +184,10 @@ export abstract class BaseToolHandler implements ToolHandler {
   getRegisteredToolConfig(runtime: ServerRuntime): ToolConfig {
     const config = this.getToolConfig();
 
-    // alwaysEnabled is an identity-checked sentinel, so bail before the
-    // per-connection scan below: connection-agnostic tools (docs, diagnostics)
-    // never route to a connection and so never need a connectionId parameter.
-    if (this.predicate === alwaysEnabled) return config;
+    // Connection-independent tools (docs, diagnostics) never route to a
+    // connection and so never need a connectionId parameter; bail before the
+    // per-connection scan below.
+    if (this.isConnectionIndependent) return config;
 
     const ids = this.enabledConnectionIds(runtime);
 
@@ -280,6 +286,18 @@ export abstract class BaseToolHandler implements ToolHandler {
   }
 
   /**
+   * Whether this tool is enabled regardless of the configured connections —
+   * true exactly when its {@linkcode predicate} is `alwaysEnabled`, the only
+   * predicate that returns enabled without inspecting a connection. Such tools
+   * (docs lookup, server diagnostics) never route to a connection, so they
+   * register even on a zero-connection config, where the per-connection verdict
+   * map driving {@linkcode enabledConnectionIds} is empty.
+   */
+  get isConnectionIndependent(): boolean {
+    return this.predicate === alwaysEnabled;
+  }
+
+  /**
    * Per-connection verdict map for this tool — the single source of truth for
    * enablement. Each verdict composes the {@linkcode predicate} with the
    * read-only overlay (see {@linkcode connectionVerdict}). Powers grouped
@@ -323,14 +341,13 @@ export abstract class BaseToolHandler implements ToolHandler {
   }
 
   /**
-   * Resolves the single connection enabled for this tool, returning the
-   * connection id, its config, and the matching client manager. Designed
-   * for handlers that look up `runtime.clientManagers[connId]` (multi-
-   * connection-ready shape).
+   * Resolves the first connection enabled for this tool, returning the
+   * connection id, its config, and the matching client manager.
    *
-   * Selects `enabledConnectionIds(runtime)[0]` — current runtime is single-
-   * connection, so this is unambiguous; if multi-connection support lands
-   * later, handlers can switch to iterating ids.
+   * Selects `enabledConnectionIds(runtime)[0]`. On a multi-connection config
+   * that arbitrarily picks the first enabled connection, which is why handlers
+   * route via {@linkcode resolveConnection} instead. Single-connection
+   * scaffolding with no remaining production callers; deletion tracked in #554.
    */
   protected resolveSoleConnection(runtime: ServerRuntime): ResolvedConnection {
     const connId = this.enabledConnectionIds(runtime)[0]!;
--- a/src/confluent/tools/base-tools.ts
+++ b/src/confluent/tools/base-tools.ts
@@ -311,12 +311,17 @@ export abstract class BaseToolHandler implements ToolHandler {
    * to launch the browser flow only when a call actually targets the OAuth
    * connection, not merely because one exists somewhere in the config.
    *
-   * Delegates to {@linkcode resolveConnection} so routing is decided in one
-   * place rather than duplicated in the gate. Returns `undefined` for a
-   * connection-independent tool (it routes to no connection) and for a call
-   * `resolveConnection` can't resolve unambiguously (e.g. a multi-connection
-   * tool invoked without a `connectionId`); in the latter case the gate declines
-   * to pre-launch a login and `handle()` reports the routing error to the caller.
+   * Returns `undefined` for a connection-independent tool — it routes to no
+   * connection, so there is nothing to pre-launch a login for.
+   *
+   * Otherwise delegates to {@linkcode resolveConnection} so routing is decided
+   * in one place rather than duplicated in the gate, and propagates its throw.
+   * Every throw path there is a `"Wacky --"` invariant violation: the injected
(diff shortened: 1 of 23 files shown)


src/server-runtime.ts
```diff
@@ -58,24 +58,38 @@ export class ServerRuntime {
...
+   * Disconnect every per-connection client manager. The teardown counterpart
+   * to the per-connection construction in {@link fromConfig}; a multi-connection
+   * config tears all of its connections down, where the single-connection
+   * {@link clientManager} getter could only reach one.
+   */
+  async disconnectAll(): Promise<void> {
+    await Promise.all(
+      Object.values(this.clientManagers).map((manager) => manager.disconnect()),
+    );
+  }
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/index.ts
```diff
@@ -448,7 +450,7 @@ async function main() {
       await transportManager.stop();
       // shutdown() is race-safe with an in-flight bootstrap.
       runtime.oauthHolder?.shutdown();
-      await runtime.clientManager.disconnect();
+      await runtime.disconnectAll();
       process.exit(0);
```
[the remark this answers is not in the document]
jlrobins: Fixed.

src/confluent/tools/base-tools.ts
```diff
@@ -279,6 +294,44 @@ export abstract class BaseToolHandler implements ToolHandler {
...
+   * map driving {@linkcode enabledConnectionIds} is empty.
+   */
+  get isConnectionIndependent(): boolean {
+    return this.predicate === alwaysEnabled;
+  }
+
+  /**
+   * The connection id this invocation routes to — the OAuth-login gate uses it
+   * to launch the browser flow only when a call actually targets the OAuth
+   * connection, not merely because one exists somewhere in the config.
```
flippingbits, line 312: Nice improvement!

src/confluent/tools/base-tools.ts
```diff
@@ -279,6 +294,44 @@ export abstract class BaseToolHandler implements ToolHandler {
...
+   */
+  resolvedTargetConnectionId(
+    runtime: ServerRuntime,
+    toolArguments: Record<string, unknown> | undefined,
+  ): string | undefined {
+    if (this.isConnectionIndependent) return undefined;
+    try {
+      return this.resolveConnection(runtime, toolArguments).connId;
+    } catch {
+      return undefined;
```
flippingbits: nit: Should we log the error instead of swallowing it?

flippingbits (approved):
I had only a single minor question. Otherwise, your PR looks good to me. I'd be happy to re-approve after fixing the merge conflict.

src/confluent/tools/base-tools.ts
jlrobins: Fixed -- should not catch this Wacky error at all --- should not happen if Zod did its job (a bad connectionId value should not have made it this far).

flippingbits (approved):
Thanks!

PR #593, opened 2026-06-13
Expose a connection's read-onlyness via  tool.

### LINKED ISSUE #589: list-configured-connections tool needs to also expose the read-only-ness of a connection.
Augment the output for `list-configured-connections` to include the read-only-ness (true or false) of each connection.

jlrobins:
# Expose connection read-only-ness in `list-configured-connections`

## `list-configured-connections` reports read-only-ness

- Each connection entry in the tool's structured payload now carries a `readOnly` boolean, so an agent can tell at a glance which connections are write-protected without having to infer it from the absence of mutating tools. The value is the connection's resolved `read_only` setting, coerced to a definite `true`/`false` (the field is Zod-defaulted to `false` on the YAML path but optional on the type).
- The text summary marks read-only connections inline with a `[read-only]` tag (e.g. `prod [read-only] (1 tool): list-topics`), keeping the at-a-glance view in sync with the authoritative structured payload.
- The tool description now advertises the `readOnly` field so agents discover it from `tools/list`.
- New handler tests pin the true and false branches in the structured payload, the `undefined → false` coercion for a connection that never set `read_only`, and the `[read-only]` summary marker; the existing exact-match assertions were extended to include `readOnly`, so any future regression that drops the field fails loudly.

## Relationships

- Closes #589
- Child of #532 (Epic: Multi-connection support)
- Builds on #578, which introduced the per-connection `read_only` flag

Screenshot 2026-06-12 at 9 38 46 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ All notable changes to this MCP server will be documented in this file.
 #### Configuration
 
 - **Per-connection `description` field.** Optional free-text label on any connection, echoed back by `list-configured-connections`.
-- **Per-connection `read_only` flag.** Set `read_only: true` on a connection to auto-disable every state-mutating tool for it, leaving only read-only tools enabled.
+- **Per-connection `read_only` flag.** Set `read_only: true` on a connection to auto-disable every state-mutating tool for it, leaving only read-only tools enabled. `list-configured-connections` reports each connection's read-onlyness.
 
 ## 1.4.0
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ All notable changes to this MCP server will be documented in this file.
 #### Configuration
 
 - **Per-connection `description` field.** Optional free-text label on any connection, echoed back by `list-configured-connections`.
-- **Per-connection `read_only` flag.** Set `read_only: true` on a connection to auto-disable every state-mutating tool for it, leaving only read-only tools enabled.
+- **Per-connection `read_only` flag.** Set `read_only: true` on a connection to auto-disable every state-mutating tool for it, leaving only read-only tools enabled. `list-configured-connections` reports each connection's read-onlyness.
```
flippingbits: Would you mind documenting the default value (`false`) here?

jlrobins: Good call -- done!

flippingbits (approved):

PR #596, opened 2026-06-13
Fix serialization of AVRO toplevel primitives.

### LINKED ISSUE #587: `produce-message` needs to be able to provide primitives as either keys or values for AVRO schemas
AVRO supports encoding of primitives like long, etc. We need to be able to _produce_ messages using primitives for key or value payloads to support faithful message formats when key or value schemas demand. Right now, can only produce with objects or string key or value payloads.

jlrobins:
# produce-message: primitive key/value payloads for Schema Registry

## Primitive payloads on the produce path

- `produce-message` could only accept an object or a string payload for a key or value. Schemas whose top-level type is an Avro (or JSON Schema) primitive — `long`, `int`, `double`, `boolean`, `string` — had no way to be produced faithfully. This widens the accepted payload to also include `number` and `boolean`, so a record keyed or valued by a primitive schema can be produced as the AI assistant naturally represents it.
- Two gates rejected primitives, and a primitive tripped both:
  - The handler's Zod `message` schema was `union([object, string])`, rejecting `number`/`boolean` before `handle()` ran.
  - `serializeMessage` threw "message must be an object matching the schema" for any non-object on the Schema Registry path — which also meant a top-level `string` schema was broken there, not just numeric ones.
- The downstream `@confluentinc/schemaregistry` serializer already handles primitive-typed schemas (`avroType.toBuffer(value)`); the only thing standing in the way was our own pre-flight validation. The fix relaxes that guard to reject only a null/undefined payload (the one value the SDK itself refuses, with a cryptic "message is empty") and lets every concrete primitive through to be validated against the real schema.

### Scope notes

- Avro `long` values beyond JavaScript's 2^53 safe-integer range lose precision because tool arguments arrive as already-parsed JSON numbers. Faithfully round-tripping large longs would require accepting them as strings plus a custom avsc long type; that is left as a follow-up rather than folded in here.
- Top-level Avro `null` is intentionally not accepted — the serializer rejects a null payload outright, so advertising it would only bait a confusing failure.

## Tests

- `serializeMessage` gains its first unit coverage: an `it.each` produces `long`/`boolean`/`string` primitive payloads against a mocked registry and round-trips each back through `deserializeMessage` to the exact value, proving the primitive reaches the serializer and encodes faithfully. The non-Schema-Registry path is pinned too (`123` → `"123"`, `true` → `"true"`), and a negative case asserts the relaxed guard still rejects a null payload with the new "a non-null message payload is required" message.
- A `produce-message` handler case produces a numeric Avro value end-to-end through `runtimeWithDecoy`, asserting success and that the producer received the serialized buffer — red today on the Zod gate.
- An integration case registers a `"long"` value schema and produces a numeric record against live Schema Registry.

## Proof
A prompt like `Copy the most recent 10 messages from ccloud RealWorldData's WeatherData to the local WeatherData topic.` now works, where said source and dest topic have key schemas specifying AVRO primitive `long` values. Here is a screenshot of VS Code Extension's consume message of one of the copied messages showing that the key was indeed serialized using AVRO:

image

## Relationships

Closes #587.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ All notable changes to this MCP server will be documented in this file.
 - **Per-connection `description` field.** Optional free-text label on any connection, echoed back by `list-configured-connections`.
 - **Per-connection `read_only` flag.** Set `read_only: true` on a connection to auto-disable every state-mutating tool for it, leaving only read-only tools enabled.
 
+### Fixed
+
+- `produce-message` can now produce primitive key and value payloads (numbers, booleans, strings) against top-level primitive Schema Registry schemas such as Avro `long`; previously the serializer rejected anything but an object.
+
 ## 1.4.0
 
 ### Added
--- a/src/confluent/schema-registry-helper.test.ts
+++ b/src/confluent/schema-registry-helper.test.ts
@@ -1,5 +1,9 @@
 import { SerdeType } from "@confluentinc/schemaregistry";
-import { checkSchemaNeeded } from "@src/confluent/schema-registry-helper.js";
+import {
+  checkSchemaNeeded,
+  deserializeMessage,
+  serializeMessage,
+} from "@src/confluent/schema-registry-helper.js";
 import { getMockedSchemaRegistry } from "@tests/stubs/index.js";
 import { beforeEach, describe, expect, it } from "vitest";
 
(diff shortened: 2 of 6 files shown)


flippingbits (approved):
LGTM

PR #604, opened 2026-06-16
New multi-connection sample configurations and tests.

### LINKED ISSUE #542: Update sample configs for multi-connections
Provide sample YAML configurations for multiple connections

jlrobins:
# Sample configs for multiple connections

## Sample configs

The recent multi-connection work shipped without a copy-pasteable starter showing it off.
These three new `sample_configs/` files each pair a **read/write local (Docker) Kafka + Schema Registry** connection with a **read-only second connection**, one per Cloud/CP authentication flavor:

- `local-docker-kafka-and-sr-and-readonly-oauth-ccloud.yaml` — local dev alongside read-only Confluent Cloud via OAuth (browser PKCE, no API keys).
- `local-docker-kafka-and-sr-and-readonly-ccloud-via-api-keys.yaml` — local dev alongside read-only Confluent Cloud via API keys, exposing `kafka` + `schema_registry` + `confluent_cloud` blocks.
- `local-docker-kafka-and-sr-and-readonly-confluent-platform.yaml` — local dev alongside a read-only self-managed Confluent Platform cluster, modeled on the existing `confluent-platform.yaml` (PLAIN/SASL_SSL, the commented SCRAM/mTLS override, internal-CA `NODE_EXTRA_CA_CERTS` note).

Each file's header comment makes the local-Docker story explicit: the Confluent for VS Code extension makes managing the local Kafka + Schema Registry containers simple, and the `localhost:9092` / `http://localhost:8081` values will vary per setup.

## Validity guard

Added a `sample_configs starters` block to `src/config/yaml-fixtures.test.ts` that discovers every `sample_configs/*.yaml` at runtime and asserts each parses through `loadConfigFromYaml` without throwing.
A `beforeEach` reads the directory into a describe-level array; the single test loops it, feeding a dummy value for each no-default `${VAR}` placeholder (defaulted placeholders keep their defaults, which are crafted to satisfy field-shape constraints like the `env-` prefix).
This mirrors the pre-existing `config.example.yaml` parse test — a shipped starter that silently stops parsing as the schema evolves is worse than no starter.
Before this change no test exercised `sample_configs/` at all; the guard now covers all seven samples and auto-covers any future addition.

## Docs

Added one line to `CONFIGURATION.md`'s "Multiple connections" section pointing at the new two-connection starters; the existing generic `sample_configs/` pointers in `README.md` and elsewhere stay as-is.

## Relationships

* Closes #542.
* Child of #532 
* Builds on the multi-connection support landed in #592 (require `connectionId` when >1 connection) and #593 (per-connection read-only exposure).

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -151,6 +151,7 @@ The fully annotated example is in [`config.example.yaml`](config.example.yaml); 
 
 A single `config.yaml` may define several named entries under `connections:` — for example a Confluent Cloud connection alongside a local Apache Kafka broker — and each tool call routes to the connection you address by id.
 There is no fixed ceiling on the number of connections.
+Ready-to-use two-connection starters pairing a read/write local-Docker connection with a read-only Cloud or Confluent Platform connection — across OAuth, API-key, and CP auth — live in [`sample_configs/`](sample_configs/).
 
 One rule constrains the mix: you may pair as many `type: direct` connections as you like, but **at most one** may be `type: oauth`.
 The shared Confluent Cloud sign-in owns a single browser session and identity, so a second OAuth connection is rejected at startup with `Multiple OAuth connections defined in configuration; only one is supported`.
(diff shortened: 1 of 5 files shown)


flippingbits (approved):
LGTM, thanks! I like that you added tests covering the sample configs.

PR #608, opened 2026-06-17
Fix coverage reports post pnpm transition.

jlrobins:
# Restore coverage generation broken by the pnpm migration

## Coverage flag no longer reaches vitest

- The pnpm migration mechanically rewrote the `*:coverage` scripts from `npm run … -- --coverage`
  to `pnpm run … -- --coverage`. The two package managers disagree about the `--` separator:
  **npm** consumes one leading `--` and forwards just `--coverage`, while **pnpm** preserves the
  `--`. So the nested call now expands to `vitest run --project unit … -- --coverage`, and vitest
  treats everything after `--` as a positional test-name filter — the `--coverage` flag is
  silently swallowed and coverage never turns on.
- Two user-visible symptoms, one root cause:
  - **Local:** `pnpm run test:unit:coverage` stopped writing the `./coverage/` HTML report.
  - **CI:** `.semaphore/semaphore.yml` runs `pnpm run test:unit:coverage` and then
    `emit-sonarqube-data`; with no `coverage/lcov.info` produced, SonarQube stopped receiving any
    coverage data.
- Fix: drop the `--` from the three `*:coverage` scripts (`test:coverage`, `test:unit:coverage`,
  `test:integration:coverage`). pnpm forwards the trailing `--coverage` to the nested script
  without it, so vitest sees the flag again. `vitest.config.ts` was never the problem — it still
  selects `lcov` under `CI` and `text`/`html`/`json-summary` locally; coverage simply never
  started, so no reporter could fire.

### Verification

- Local: `pnpm run test:unit:coverage` → `Coverage enabled with v8`; `coverage/index.html`,
  `coverage-summary.json`, and the per-file `src/` HTML are regenerated.
- CI path: `CI=1 pnpm run test:unit:coverage` → `coverage/lcov.info` is produced (the artifact
  `emit-sonarqube-data` uploads).

## Relationships

Fixes regression introduced by #583 (`build: migrate package manager from npm to pnpm`).


flippingbits (approved):
Thanks! Let's give it a try.

PR #609, opened 2026-06-17
Additional ListSchemasHandler.handle() tests.

jlrobins:
# Targeted tests for `ListSchemasHandler.handle()`

## Unit coverage

`ListSchemasHandler.handle()` previously had only the `latestOnly=true` happy path, the prefix filter, `environment_id` wiring, and the top-level `getAllSubjects`-rejects case under test — 58% line / 25% branch. The handler's `latestOnly=false` versions branch and every per-item error sub-branch were untested, and `getToolConfig()` had no test at all.

- **`latestOnly=false` versions branch.** New cases prove each subject maps to an array built from `getAllVersions` + per-version `getSchemaMetadata`, that `getLatestSchemaMetadata` is never consulted, and that `deleted` defaults to `false (and forwards `true` when supplied) into `getSchemaMetadata`.
- **Per-item error resilience.** Separate cases pin that a rejection from `getLatestSchemaMetadata` (latest mode), from `getSchemaMetadata` (one version), or from `getAllVersions` (whole subject) is captured as a per-subject / per-version `{ error }` entry rather than failing the whole call — asserting the exact error string and surrounding shape.
- **Non-`Error` rejection fallbacks.** The SR SDK can reject with a non-`Error` value; four cases drive each `catch` block's `err instanceof Error ? … : …` fallback arm (`String(err)` inside the loops, `JSON.stringify` at the top level), closing the branch gap to 100%.
- **`getToolConfig()`.** Pins the tool name, `READ_ONLY` annotations by identity, and the four input-schema field names.

Result: 100% statement / branch / line / function coverage on the handler.

## Integration coverage

The integration suite previously exercised only the default `latestOnly=true` path end-to-end. Added a direct-connection case calling with `latestOnly: false`, asserting the seeded subject maps to a one-element version
array (version 1, numeric id) — driving the server-side `getAllVersions` + per-version `getSchemaMetadata` branch against a live registry.

## Relationships

Backfills missing coverage for `ListSchemasHandler` — no issue tie.

### DIFF
--- a/src/confluent/tools/handlers/schema/list-schemas-handler.integration.test.ts
+++ b/src/confluent/tools/handlers/schema/list-schemas-handler.integration.test.ts
@@ -92,6 +92,23 @@ describe(
           const parsed = JSON.parse(textContent(result));
           expect(parsed).toHaveProperty(subject);
         });
+
+        it("should return an array of every version for the subject when latestOnly is false", async () => {
+          const result = await server.client.callTool({
+            name: ToolName.LIST_SCHEMAS,
+            arguments: { subjectPrefix: subject, latestOnly: false },
+          });
+
+          expect(result.isError, textContent(result)).not.toBe(true);
+          // with latestOnly:false the handler walks getAllVersions + getSchemaMetadata per
+          // version, so each subject maps to an array rather than a single metadata object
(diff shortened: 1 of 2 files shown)


flippingbits (approved):
Nice improvement!

PR #610, opened 2026-06-17
Remove the deprecated FLINK_ENV_NAME env var.

### LINKED ISSUE #598: Remove support for env var FLINK_ENV_NAME.
Should have gone away in 1.4.0 timeline. See comments in src/env-schema.ts.

jlrobins:
## `FLINK_ENV_NAME` removed

- The legacy `FLINK_ENV_NAME` environment variable is gone. It was a deprecated alias of `FLINK_CATALOG_NAME` that originally targeted removal in the v1.4.0 timeline; that window came and went, so it's removed now. Users still on the old spelling must rename to `FLINK_CATALOG_NAME` — or, on the YAML path, set a connection's `flink.catalog_name` (same meaning — the Flink catalog name used as `sql.current-catalog`).
- Dropped the field from `envSchema`, so the variable is no longer recognized at the env boundary — a raw `process.env` carrying `FLINK_ENV_NAME` now silently strips it rather than mapping it to `catalog_name`.
- `resolveFlinkCatalogName()` collapsed away: with no alias to reconcile, catalog-name resolution is just a direct read of `FLINK_CATALOG_NAME`. The "both set" conflict error and the deprecation warning are deleted with it.
- An `env-schema.test.ts` case pins the new boundary behavior: a raw env carrying `FLINK_ENV_NAME` parses without surfacing it on the result, which would have failed while the field still existed.

## Docs

- `CONFIGURATION.md` drops the `FLINK_ENV_NAME` table row; `.env.example` drops the deprecation comment; `CHANGELOG.md` records the removal under Unreleased.

## Relationships

Closes #598


flippingbits (approved):
🎉

PR #611, opened 2026-06-17
`create-topics` support for `replicationFactor`

### LINKED ISSUE #585: `create-topics` tool needs to offer `replicationFactor`
Users need to be able to specify `replicationFactor` when creating topics. While Confluent Cloud-based brokers default to and will only accept a RF of 3, onprem and local clusters will vary.

jlrobins:
# `create-topics`: optional `replicationFactor`

## Replication factor support

- `create-topics` now accepts an optional per-topic `replicationFactor` alongside the existing `numPartitions`. The underlying `@confluentinc/kafka-javascript` `ITopicConfig` already carried `replicationFactor?: number`; the handler simply wasn't forwarding it — its `.map()` projected each topic down to `{ topic, numPartitions }` and silently dropped everything else. The fix adds the Zod field and threads it through the map.
- The field documents the deployment reality: Confluent Cloud clusters only ever support a replication factor of 3 (the broker rejects any other value at creation), so callers should pass 3 or omit the field; on-premises or other clusters may accept other values. This was confirmed empirically — an integration attempt at RF=4 came back `Topic replication factor must be 3`.
- The tool description and the captured sample listing in `README.md` were updated to mention replication factor so the docs don't lag the schema.

## Tests

- A unit test pins the forwarding with a literal assertion on the `admin.createTopics` call shape (`{ timeout, topics: [{ topic, numPartitions, replicationFactor }] }`). It was written red first: against the pre-fix `.map()` the key was stripped and the call omitted it. A companion case asserts the key stays absent when the caller omits it, so the broker-default path is protected.
- A direct-connection integration case creates a topic with `replicationFactor: 3` and asserts the call succeeds and the topic appears, proving the explicit-RF payload round-trips to and is accepted by the broker. It deliberately does not assert a non-default realized RF — CCloud permits no other value, so realized RF can't distinguish "honored" from "default"; the unit test carries that proof. The CP single-broker suite is left untouched since any RF > 1 fails there for want of brokers.

## Relationships

Closes #585.


src/confluent/tools/handlers/kafka/create-topics-handler.ts
```diff
@@ -67,10 +77,13 @@ export class CreateTopicsHandler extends BaseToolHandler {
...
-            topic,
-            numPartitions,
-          })),
+          topics: parsed.topics.map(
+            ({ topic, numPartitions, replicationFactor }) => ({
+              topic,
+              numPartitions,
+              replicationFactor,
+            }),
+          ),
```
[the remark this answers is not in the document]
jlrobins: Good instinct to poke at the `undefined`-on-the-wire seam, but both halves of this check out as already-handled here:

- **The omit test passes.** `toHaveBeenCalledWith` uses `toEqual` semantics, and `toEqual` treats `{ replicationFactor: undefined }` as equal to a missing key — so the "omit" assertion is green (verified in the full unit run). It still bites: a buggy `replicationFactor: 5` would fail it.
- **The broker accepts `undefined` RF.** This is symmetric with the pre-existing `numPartitions: undefined` the map already passes when partitions are omitted (the "broker default" contract). Confirmed live: the plain-create and OAuth-create integration cases both route through the new map sending `replicationFactor: undefined` and create the topic successfully.

Conditionalizing only `replicationFactor` would also make it inconsistent with `numPartitions` in the same literal (which stays unconditional), so leaving both symmetric is the smaller, tested change. No code change on this one.

src/confluent/tools/handlers/kafka/create-topics-handler.ts
```diff
@@ -28,6 +28,16 @@ const createTopicArgs = z.object({
...
           ),
+        replicationFactor: z
+          .number()
+          .int()
+          .positive()
+          .optional()
+          .describe(
+            "Replication factor for the topic. If not specified, the broker default is used. " +
+              "Confluent Cloud clusters only ever support a value of 3 (or omit this to take the default); " +
+              "on-premises or other clusters may accept other values.",
```
flippingbits: nit:


```suggestion
              "Confluent Platform or other clusters may accept other values.",
```

flippingbits (commented):
I have only a very minor comment on the argument description. Otherwise, your PR looks good to me.

src/confluent/tools/handlers/kafka/create-topics-handler.ts
jlrobins: Reworded, thanks!

flippingbits (approved):
Thanks!

Issue #534, opened 2026-06-03
New final protected method BaseToolHandler.resolveConnection()


jlrobins:
New method to ultimately replace `resolveSoleConnection()`:

```typescript
/**
 * Resolves which connection a tool call targets: the explicit `connectionId`
 * argument when present, else the sole connection enabled for this tool.
 * Throws a listing error when the id is unknown/not-enabled, or when omitted
 * while two or more connections are candidates (the augmented schema makes the
 * latter unreachable in normal MCP flow — this is the defensive backstop).
 */
protected resolveConnection(
  runtime: ServerRuntime,
  toolArguments: Record<string, unknown> | undefined,
): { connId: string; conn: ConnectionConfig; clientManager: BaseClientManager };
```

It reads `connectionId` off the **raw** `toolArguments` — the framework already validated it against the registered schema — rather than off the handler's locally-parsed object, whose `z.object` re-`parse()` would strip the unknown key.
So handlers never carry `connectionId` in their own Zod schema; the base class owns the param on both the publish and consume sides. (See #533)

direct-or-oauth-neutral Tools will ultimately call this at the top of their handle() in lieu of their call to `resolveSoleConnection()`.

Code up the method and a complete unit test suite as thorough as was done for #533.

Issue #151, opened 2026-04-10
Migration to YAML configuration of a single connection


shouples:
Wean off of env-var driven configuration in favor of a YAML document format describing a list of connections with names and types and additional information ultimately representing all of the possible env vars. This first milestone will only allow / accept a single connection defined in the list and the resulting configuration will be used to configure / construct the `DefaultClientManager` directly instead of from env vars explicitly.

Because we have already passed 1.0 release, will probably need to have one or two releases where we will support either env var OR yaml configuration before removing 'only indirect env var' configuration.

A future epic issue will then expand the capabilities to allow for more than one connection (and of varying types) to be defined, then used to construct something other than the DefaultClientManager object, but instead something which can cleanly contain and represent multiple connections.

Base YAML off of the good bits at https://confluentinc.atlassian.net/wiki/spaces/DTX/pages/5405148158/Migration+to+YAML-based+MCP+Server+Configuration

PR #618, opened 2026-06-22
Delete the single-connection scaffolding accessors

### LINKED ISSUE #554: Delete the single-connection scaffolding accessors

jlrobins:
## Delete the single-connection scaffolding accessors

- Removed the four "sole/singular" accessors that the multi-connection epic drove to zero production callers: `MCPServerConfiguration.getSoleConnection()` / `getSoleDirectConnection()`, `BaseToolHandler.resolveSoleConnection()`, and the singular `ServerRuntime.clientManager` getter. These encoded the "there is only one connection" assumption; #534/#538/#539/#541 routed every handler through explicit `connectionId` resolution and #540 dropped the parse-time guard, leaving these as dead code whose deletion this issue owns. Behavior-preserving: no runtime path changes.
- Migrated every remaining caller — all of them tests and harness helpers — off the deleted accessors. The replacements are uniformly **by-id addressing plus narrow-to-direct**, never "find the sole one", which is the whole point: a test fixture that holds one connection is addressed by that connection's known id, not by an accessor that assumes the count.

## Test/harness migration

- Standardized the single-connection fixtures onto a consistent `default` connection id — both the external `test-fixtures/yaml_configs/valid/*.yaml` files (a mix of `production` / `local`) and the inline single-connection YAML strings in `index.test.ts` — matching the id `tests/factories/runtime.ts` and the existing OAuth fixture already use. The config-parser/builder unit tests now read their one connection via `getConnectionConfig("default")` rather than a sole-lookup. The deliberate multi-connection / negative cases keep their distinct ids.
- Added `directConnectionOf(config, id)` (`tests/factories/config.ts`) and `integrationDirectConnection()` (`tests/harness/runtime.ts`) as the narrow-to-direct seam over `getConnectionConfig(id)`, which returns the `ConnectionConfig` union. These replace `getSoleDirectConnection()` at the config-test and integration-harness sites respectively. The `getSoleConnection()` (union) sites migrated straight to `getConnectionConfig(id)`. Migrating the integration tests off the accessor also orphaned their `const runtime = integrationRuntime()` locals (the accessor was their only consumer), so those and the now-unused `integrationRuntime` imports came out too.
- Deleted the now-orphaned tests that exercised the accessors directly (`getSoleConnection` / `getSoleDirectConnection` blocks in `models.test.ts`, `resolveSoleConnection()` in `base-tools.test.ts`, `get clientManager()` in `server-runtime.test.ts`). The surviving `getConnectionConfig` coverage is the by-id replacement.
- Swept the prose that narrated the deleted scaffolding as live: the `disconnectAll` docstring's back-reference to the singular getter, four `getSoleConnection()`-citing docstrings in the integration harness, the `runtimeWithDecoy` docstring and its `.claude/rules/unit-tests.md` echo (both described the failure mode as "the legacy sole-connection accessor" — now stated as the actual `enabledConnectionIds[0]` anti-pattern). Several "stale prose" targets the issue lists were already clean on this branch (the predicate docstrings, and server-runtime's `#151` / `enforceSingleConnectionOnly()` references).

## Relationships

- Closes #554.
- Child of #532 (Epic: Multi-connection support).

### DIFF
--- a/.claude/rules/unit-tests.md
+++ b/.claude/rules/unit-tests.md
@@ -393,7 +393,7 @@ A few case kinds legitimately stay as direct `handler.handle(...)` calls rather 
 
 Such cases don't carry decoy coverage; the suite's other (harness-routed) cases provide it.
 
-A reverted handler (one that goes back to `resolveSoleConnection()`) trips the decoy's untouched assertion on every routed case, so the coverage is not vacuous.
+A handler that resolves by grabbing `enabledConnectionIds[0]` instead of the caller's `connectionId` trips the decoy's untouched assertion on every routed case, so the coverage is not vacuous.
 
 ## Test Server & Runtime Fixtures
 
--- a/src/config/env-config.test.ts
+++ b/src/config/env-config.test.ts
@@ -2,6 +2,7 @@ import {
   buildConfigFromEnvAndCli,
   DEFAULT_CONNECTION_ID,
 } from "@src/config/env-config.js";
+import { directConnectionOf } from "@tests/factories/config.js";
 import { describe, expect, it } from "vitest";
 
 type EnvInput = Parameters<typeof buildConfigFromEnvAndCli>[0];
@@ -63,8 +64,7 @@ describe("config/env-config.ts", () => {
           }),
         );
 
-        const conn = config.getSoleDirectConnection();
-        expect(conn.type).toBe("direct");
+        const conn = directConnectionOf(config, DEFAULT_CONNECTION_ID);
         expect(conn.kafka?.bootstrap_servers).toBe("localhost:9092");
         expect(conn.schema_registry?.endpoint).toBe("http://localhost:8081");
       });
@@ -84,7 +84,7 @@ describe("config/env-config.ts", () => {
           envWith({ BOOTSTRAP_SERVERS: "broker1:9092,broker2:9092" }),
         );
 
-        const conn = config.getSoleDirectConnection();
+        const conn = directConnectionOf(config, DEFAULT_CONNECTION_ID);
         expect(conn.kafka?.bootstrap_servers).toBe("broker1:9092,broker2:9092");
         expect(conn.schema_registry).toBeUndefined();
       });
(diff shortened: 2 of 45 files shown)


bcscc (approved):
Satisfying clean-up thanks!

PR #628, opened 2026-06-23
Deflake the oauth integration tests.

jlrobins:
# Deflake the kafka + OAuth integration suite

## The flake

The `@kafka` CI job intermittently failed with `Hook timed out in 180000ms` in an OAuth describe's `beforeAll` at `startOAuthServer` — a different test each run (`delete-topics` oauth/stdio one run, `create-topics` oauth/http the next). Root cause: every OAuth describe must bind a single hard-coded Auth0 callback port (26640), so they all serialize through one filesystem lock (`acquireOAuthPortLock`), acquired _inside_ the `beforeAll`. With 11 kafka files each running an OAuth describe across 3 transports (~33 serialized lock-holds, each spanning a real Playwright→Auth0 sign-in), and `pool: "forks"` starting ~8 files at once, the forks queued on the lock. A `beforeAll`'s 180s timeout includes the time blocked waiting for the lock, so a fork stuck behind enough holders died before doing any work. The lock's own 300s timeout never fired because the shorter hook timeout killed the hook first — yielding the opaque message instead of naming the holder.

## The fixes

Three changes, smallest-blast-radius first:

- **OAuth describes run on stdio only.** New `activeOAuthTransports` (stdio-filtered `activeTransports`) in `tests/harness/transports.ts`; **all 19 OAuth describes** across the suite now iterate it instead of `activeTransports` (direct describes keep all three; the OAuth-flow smoke test was already stdio-only). The Auth0/PKCE sign-in is transport-agnostic — the browser flow is identical regardless of MCP transport, and the transport layer is already covered by the parallel direct describes and the smoke suite. This alone cuts each OAuth job's serialized lock-holds ~3× (kafka: 33 → 11). The `.claude/rules/integration-tests.md` dual-mode pattern now documents `activeOAuthTransports` as the rule for OAuth describes.
- **Lock timeout below the hook timeout.** `acquireOAuthPortLock` default dropped from 300s to 150s (< the 180s OAuth `beforeAll`). If a holder ever genuinely wedges, the failure now reads `lock not released within … (holder PID=…)` and names the culprit, rather than the opaque `Hook timed out`.
- **OAuth lane runs sequentially.** The Makefile appends `--no-file-parallelism` when `INTEGRATION_TEST_CONNECTION_TYPE=oauth`, and the per-PR `@kafka` block is split into two parallel jobs (separate agents): `@kafka direct` (parallel) and `@kafka oauth` (sequential). With the OAuth lane on its own sequential agent, the callback-port lock is never contended — each file acquires it instantly — so no `beforeAll` ever sits in a queue. The direct lane keeps full parallelism, and each lane publishes its own `TEST-integration.xml`.

## Wall clock

Unaffected-to-better. The OAuth work was _already_ serialized by the lock, so forcing the OAuth lane sequential doesn't slow it — it just stops the other forks from busy-waiting and tripping the 180s timeout. The direct lane stays parallel (separate job). And stdio-only OAuth actually cuts OAuth wall time ~3×. Net: faster, and no more timeout-driven retries.

## Note: scheduled full run

This fixes the per-PR path (where the flake was observed). The scheduled `integration.yml` run with `CONNECTION_TYPE=all` still executes kafka direct+oauth in one combined, parallel invocation; stdio-only OAuth makes a pile-up ~3× less likely there, but it isn't split. Running that pipeline with `CONNECTION_TYPE=oauth` already serializes via the Makefile knob. Splitting the scheduled kafka cell too is a small follow-up if the daily run proves flaky.

## References
A sample doomed CICD is over at https://semaphore.ci.confluent.io/branches/f20ded17-2c09-416b-9b22-eca8f226dc30

### DIFF
--- a/.claude/rules/integration-tests.md
+++ b/.claude/rules/integration-tests.md
@@ -448,13 +448,19 @@ describe("<handler>", { tags: [Tag.<GROUP>] }, () => {
       return;
     }
 
-    describe.each(activeTransports)("via %s transport", (transport) => {
+    describe.each(activeOAuthTransports)("via %s transport", (transport) => {
       // oauth-only beforeAll (startOAuthServer + 180_000 timeout) / afterAll (stopOAuthServer)
       // tool calls dispatch via callToolWithOAuthFlow(server, credentials, ...)
     });
   });
 });
 ```
+
+**Every OAuth describe iterates `activeOAuthTransports` (stdio only), not `activeTransports`** — the direct describe keeps `activeTransports` (all three).
+The Auth0/PKCE sign-in is transport-agnostic, so running it on every transport adds no coverage the direct describes and the smoke suite don't already provide; it only multiplies work.
+And that work is serialized: every OAuth describe blocks on a single hard-coded callback port (`acquireOAuthPortLock`), so ×3 transports triples the serialized lock-holds — enough, in the OAuth-heavy `@kafka` job, to push queued `beforeAll`s past their 180s hook timeout.
+The OAuth-flow smoke test (`ccloud-oauth.integration.test.ts`) is stdio-only for the same reason.
+CI additionally runs the OAuth lane sequentially; see the CI Matrix section.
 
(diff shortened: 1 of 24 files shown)


bcscc (approved):
Thanks!

PR #629, opened 2026-06-24
feat: add search-messages tool for full-text search over Kafka topics

### LINKED ISSUE #476: Add `search-topic-messages` tool for full-text search against Kafka topic messages
## Summary

Extend the local MCP server with a new tool that performs full-text search against messages in one or more Kafka topics. The tool should mirror the signature of the existing `consume-messages` tool ([src/confluent/tools/handlers/kafka/consume-kafka-messages-handler.ts:47-82](https://github.com/confluentinc/mcp-confluent/blob/main/src/confluent/tools/handlers/kafka/consume-kafka-messages-handler.ts#L47-L82)), adding a search-query argument and returning only messages whose deserialized key/value/headers match.

## Motivation

Today, the only way to find a specific message in a topic via the MCP server is to call `consume-messages` and have the AI assistant scan the returned batch. That doesn't scale beyond `maxMessages` (default 10) and forces the assistant to repeatedly consume + filter. A first-class search tool lets the assistant push the filter down into a single tool call and surface only relevant matches, with much higher recall over the scanned window.

Typical use cases:
- "Find messages in `orders` that mention customer ID `cust-42`."
- "Show me any message in `audit-log` containing the string `permission denied`."
- "Search `events` for messages where the value matches `/checkout.*failed/i`."

## Proposed Tool Shape

New tool `search-topic-messages` (`ToolName.SEARCH_TOPIC_MESSAGES`), handler at `src/confluent/tools/handlers/kafka/search-topic-messages-handler.ts`. Predicate: `kafkaBootstrapOrOAuth` (same as `consume-messages`). Annotation: `READ_ONLY`.

Arguments (mostly identical to `consumeKafkaMessagesArgs`):

| Argument | Type | Notes |
|---|---|---|
| `topicNames` | `string[]` (nonempty) | Same as consume. |
| `query` | `string` | **New.** The search term. |
| `queryMode` | `"substring" \| "regex"` (default `"substring"`) | **New.** Substring is case-insensitive by default; regex compiles `query` as a `RegExp`. |
| `searchIn` | `Array<"key" \| "value" \| "headers">` (default `["value"]`) | **New.** Which parts of each message to match against. |
| `maxMatches` | `number` (default 10) | Replaces `maxMessages` — stop once this many *matches* are found. |
| `maxScanned` | `number` (default 1000) | **New.** Upper bound on messages scanned, even if no match — avoids runaway scans on busy topics. |
| `timeoutMs` | `number` (default 10000) | Same as consume. |
| `value` | `valueOptions` | Same as consume (Schema Registry deserialization options). |
| `key` | `keyOptions` (optional) | Same as consume. |
| `cluster_id` / `environment_id` | `string` (optional) | Same as consume. |

Response: same `ProcessedMessage[]` shape returned by `consume-messages`, filtered to matches, plus a summary line like `Found N matches in M scanned messages from topics ...`.

## Implementation Notes

- Reuse the consumer setup, deserialization path (`processMessage`), and Schema Registry plumbing from `ConsumeKafkaMessagesHandler` — ideally by extracting a small shared helper rather than duplicating. Worth considering whether `ConsumeKafkaMessagesHandler.processMessage` should move to a sibling module so both handlers can import it without one depending on the other.
- Matching runs *after* deserialization so callers can search the decoded string representation of Avro/JSON/Protobuf payloads, not raw bytes.
- For `searchIn: ["headers"]`, match against both header names and string-coerced header values.
- Termination condition is `matches.length >= maxMatches || scanned >= maxScanned || timeoutMs elapsed`.
- Regex mode should reject invalid patterns up front with a clear error (don't crash mid-consume).

flippingbits:
## Summary

Implements #476. Adds a `search-messages` tool that pushes message filtering down into a single tool call instead of forcing the assistant to repeatedly `consume-messages` + scan the returned batch.

The tool scans up to `maxScanned` messages from earliest across every partition of the requested topic(s) and returns only the (up to `maxMatches`) records whose **decoded** key/value/headers match the query — case-insensitive substring by default, or a compiled regex when `queryMode: "regex"`.

It mirrors `consume-messages`:
- Same `kafkaBootstrapOrOAuth` predicate and `READ_ONLY` annotation.
- Same Schema Registry deserialization path (AVRO/JSON/PROTOBUF) with raw-bytes fallback.
- Per-invocation unique consumer group id; `topicNames` is deduplicated before subscribing.
- Matching runs **after** deserialization, so callers search the decoded string representation, not raw bytes. Invalid regex patterns are rejected up front with a clear error rather than crashing mid-consume.

Partitions are scanned concurrently (`partitionConcurrency`, default `4`) via `eachBatch` to overlap Schema Registry decode latency on wide topics.

Termination condition: the search resolves when one of five exit conditions wins — `matches.length >= maxMatches`, `scanned >= maxScanned`, every partition drained to its start-of-search high-water mark (end-of-log), `timeoutMs` elapsed, or `consumer.run()` rejects. The end-of-log exit means a search no longer waits out the full `timeoutMs` on a topic it has already scanned to the end. Because concurrent partition workers can each clear the bound check before `accepting` flips, the response is trimmed to exactly `maxMatches`. Response is the filtered `ProcessedMessage[]` plus a `Found N matches in M scanned messages from topics ...` summary line.

### Arguments

| Argument | Type | Default | Notes |
|---|---|---|---|
| `topicNames` | `string[]` (nonempty) | — | Topics to search; every partition is scanned. Deduplicated before subscribing. |
| `query` | `string` | — | The search term. |
| `queryMode` | `"substring" \| "regex"` | `"substring"` | Substring is case-insensitive; regex compiles `query` as a `RegExp`, accepting either a bare pattern (`checkout.*failed`) or regex-literal syntax with flags (`/checkout.*failed/i`). |
| `searchIn` | `Array<"key" \| "value" \| "headers">` | `["value"]` | For `headers`, matches both header names and string-coerced values. |
| `maxMatches` | `number` | `10` | Stop once this many matches are found. |
| `maxScanned` | `number` | `1000` | Upper bound on messages scanned — guards against runaway scans. |
| `timeoutMs` | `number` | `10000` | Max time to spend scanning. |
| `partitionConcurrency` | `number` (optional) | `4` | How many partitions to scan in parallel; capped at `32`. Overlaps Schema Registry decode latency across partitions for faster wide-topic searches. Values above the topic's partition count have no extra effect. |
| `valueFormat` / `keyFormat` | SR options | auto-decode | Same as `consume-messages`. |
| `cluster_id` / `environment_id` | `string` (optional) | — | Same as `consume-messages`. |

## Shared extraction

Per the issue's note, the shared deserialization plumbing (`processMessage`, `ProcessedMessage`, `formatMessageTimestamp`, schema-registry options, header echoing) moves to a new `message-processing.ts` sibling module so both handlers import it without one depending on the other. `ConsumeKafkaMessagesHandler.processMessage` now delegates and re-exports the symbols its colocated test imports, so that test is unchanged.

The shared `processMessage` also gained an optional per-invocation `SchemaLookupCache`: `search-messages` passes one so scanning up to `maxScanned` records collapses the latest-schema lookup to one `getLatestSchemaMetadata` round-trip per subject instead of one per record per side. The cache is scoped to a single `handle()` call, so a schema change between invocations is always picked up. `consume-messages` is unchanged — it doesn't pass a cache.

It also accepts an optional `DeserializerCache` (in `schema-registry-helper.ts`): `search-messages` reuses one deserializer per `(schemaType, serdeType)` for the life of a search, keeping the parsed-schema cache warm across records instead of re-compiling the writer schema every message. Like the schema-lookup cache, it's scoped to a single `handle()` call; `consume-messages` doesn't pass one.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run build`
- [x] `pnpm run test:unit`
- New `search-messages-handler.test.ts`: schema defaults/validation, `buildMatcher` (substring/bare regex/regex-literal flags/invalid pattern/invalid flag), `messageMatches` (value/key/headers/tombstone/object/BigInt), and `handle()` (invalid-regex rejection, earliest + per-call group id, topicNames dedup, match filtering + summary, maxMatches/maxScanned stops, run-rejection error).
- Test pins updated: `tool-registry.test.ts` (`EXPECTED_PREDICATES` + smoke), `index.test.ts` (OAuth-enabled), `cp-gating.test.ts` (CP-enabled).
- README: both tool tables + the `--list-tools` example output.


CLAUDE.md
```diff
@@ -77,6 +77,15 @@ Detailed conventions (handler structure, input schema rules, registration checkl
 3. Implement `getToolConfig()` (name, description, Zod input schema, `annotations`) and `handle()`.
 4. Register the handler in the `ToolHandlerRegistry.handlers` map in `src/confluent/tools/tool-registry.ts`.
 5. If the tool calls a new Confluent Cloud REST endpoint, add it to `openapi.json` and regenerate types with `pnpm run generate:openapi-types`. Commit the updated `src/confluent/openapi-schema.d.ts` alongside the `openapi.json` change.
+6. Add a `CHANGELOG.md` entry under `## Unreleased → #### New Tools / Tool Features` — see [Changelog](#changelog) below.
```
flippingbits, line 80: Not really related to this PR but too many recent Claude-generated PRs miss changelog updates, so hopefully this saves us some time.

src/confluent/tools/handlers/kafka/consume-kafka-messages-handler.test.ts
```diff
@@ -1926,6 +1926,8 @@ describe("consume-kafka-messages-handler.ts", () => {
             fakeRegistry,
             SerdeType.VALUE,
             undefined,
+            // deserializerCache: consume doesn't pass one (search does).
+            undefined,
```
flippingbits, line 1930: Maybe introduce it to the `consume-messages` tool in a follow-up?

src/confluent/tools/handlers/kafka/consume-kafka-messages-handler.ts
```diff
@@ -34,39 +35,13 @@ import { logger } from "@src/logger.js";
...
-  .default({ disableSchemaRegistry: false });
-
-// `ValueOptions` and `KeyOptions` are structurally identical to
-// `schemaRegistryOptions` — they exist as named types only to make the
-// value-side vs key-side distinction explicit at `processMessage`'s
-// signature. The schema fields below use `schemaRegistryOptions` directly;
-// no separate runtime schema is needed for each side.
-type ValueOptions = z.infer<typeof schemaRegistryOptions>;
-type KeyOptions = z.infer<typeof schemaRegistryOptions>;
+// Re-exported so existing importers (and the colocated test) can keep
```
Cerchie, line 38: Why not update the importers instead?

flippingbits: Fair question. I didn't want to touch the tests when updating this tool handler. I'd be happy to do so if you like, though.

Cerchie: no that's ok, I get that the size is large enough, maybe in a follow-on

Cerchie:
I haven't had time to read all the files deeply yet, but in terms of clicktesting, I am unable to get results for `spotify-listening-data`. It always returns 0 results, even when I query for a substring that is present in every message value. I have tried headers, keys, quotes around the search term, increasing the max-scanned value-- no luck.

Screenshot 2026-07-01 at 7 32 12 AM

flippingbits:
> I haven't had time to read all the files deeply yet, but in terms of clicktesting, I am unable to get results for spotify-listening-data. It always returns 0 results, even when I query for a substring that is present in every message value. I have tried headers, keys, quotes around the search term, increasing the max-scanned value-- no luck.

Can you share the tool response? It works for me with the given topic and search string.

Cerchie:
> > I haven't had time to read all the files deeply yet, but in terms of clicktesting, I am unable to get results for spotify-listening-data. It always returns 0 results, even when I query for a substring that is present in every message value. I have tried headers, keys, quotes around the search term, increasing the max-scanned value-- no luck.
> 
> Can you share the tool response? It works for me with the given topic and search string.

I get tool success from the response,

```
{"level":"info","time":"2026-07-06T15:16:39.878Z","pid":84280,"hostname":"REDACTED","name":"mcp-confluent","namespace":"kafka","args":[{"name":"mcp-confluent#consumer-2","fac":"BINDING","timestamp":1783350999877}],"msg":"Consumer disconnected"}
```

 but it's 0 results, which feels wrong:
```
Found 0 matches in 0 scanned messages from topics spotify-listening-data.
Matches: []
```

Cerchie:
About the label: accidentally clicked "resolve conflicts" on ReviewHub 
Screenshot 2026-07-28 at 7 23 35 AM

flippingbits:
Closing due to staleness

PR #639, opened 2026-06-26
Derive OAuth CCloud REST URLs from one base domain

bcscc:
## Summary of Changes

Infrastructure-only refactor — **0 tools enabled**, no behavior change. This is the shared URL-wiring groundwork for Telemetry / Metrics tool OAuth enablement, which builds the telemetry endpoint + data-plane-token wiring on top of the base-domain map landed here and enables `query-metrics` + `list-metrics`.

- Collapse the per-environment OAuth URL configuration to a single base-domain map and derive the Confluent Cloud REST surfaces from it.
  - `auth0-config.ts` previously stored `apiUrl` and `cloudRestUrl` as independent literal strings per environment. They are the same base domain with a different prefix, so they now derive from one `CCLOUD_DOMAINS` record:
    - `getApiUrlForEnv(env)` → `https://${D}` (session/login API host; no `api.` prefix) — new helper
    - `getCloudRestUrlForEnv(env)` → `https://api.${D}` — unchanged signature, now derived
  - Login identity (`clientId`, `domain`, `callbackUrl`, `scopes`) stays literal — login lives on a different domain family (`login.confluent.io` / `login-stag.confluent-dev.io`) and does not derive from the cloud base domain.
  - The map is the single source of truth for the surfaces (telemetry: `api.telemetry.${D}`) and (regional Flink: `flink.<region>.<cloud>.${D}`) will derive next — those helpers land with their respective PRs, not here.
- Drop the now-callerless `cloudRestUrl` field from the `Auth0Config` interface. `apiUrl` is retained — the token-exchange chain (`auth-context.ts`, `token-chain.ts`) still reads it for `/api/sessions` and `/api/access_tokens`.

This PR enables no tools and changes no runtime behavior, so it has no tool-level walkthrough. Verify the refactor instead:

1. Run `pnpm run typecheck && pnpm run test:unit && pnpm run lint` — all green (the pre-existing `schema-registry-helper.test.ts` protobuf-decode failure on `main` is unrelated).
2. New unit coverage asserts the `getApiUrlForEnv` and `getCloudRestUrlForEnv` derivations for `devel` / `stag` / `prod`; the existing `getAuth0Config` assertions (now reading derived `apiUrl`) stay green, proving the collapse preserves the exact prior URLs.

- Before: `apiUrl` and `cloudRestUrl` were independent per-env literals; adding the telemetry host (item 5) and the regional Flink base (item 6) would have meant three-to-five hand-maintained strings per environment, most of them internal-dev-only and unverifiable.
- After: one base-domain map per environment; each CCloud REST URL derives from it. Telemetry and Flink URL derivation become one-liners in their own PRs.

#### Tests

- [x] Added new
- [x] Updated existing
- [ ] Deleted existing

#### Release notes

- [ ] Does anything in this PR need to be mentioned in the user-facing CHANGELOG? — No; this PR enables no tools and changes no behavior. The user-facing "Telemetry tools now work under OAuth" note ships with the item 5 PR.

### DIFF
--- a/src/confluent/oauth/auth0-config.test.ts
+++ b/src/confluent/oauth/auth0-config.test.ts
@@ -1,4 +1,5 @@
 import {
+  getApiUrlForEnv,
   getAuth0Config,
   getCloudRestUrlForEnv,
   OAUTH_CALLBACK_PATH,
@@ -44,6 +45,20 @@ describe("oauth/auth0-config.ts", () => {
 
       expect(new Set(callbacks).size).toBe(1);
       expect(callbacks[0]).toContain("127.0.0.1:26640");
+    });
+  });
+
+  describe("getApiUrlForEnv", () => {
+    it("should return https://devel.cpdev.cloud for devel", () => {
+      expect(getApiUrlForEnv("devel")).toBe("https://devel.cpdev.cloud");
+    });
+
+    it("should return https://stag.cpdev.cloud for stag", () => {
+      expect(getApiUrlForEnv("stag")).toBe("https://stag.cpdev.cloud");
+    });
+
+    it("should return https://confluent.cloud for prod", () => {
+      expect(getApiUrlForEnv("prod")).toBe("https://confluent.cloud");
     });
   });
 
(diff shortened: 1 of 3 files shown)


eguicf (approved):
Collapsing the per-env `apiUrl/cloudRestUrl` literals into a single `CCLOUD_DOMAINS` is clear and cleaner; Tests all green. LGTM!

PR #671, opened 2026-06-30
[tests]: add applyJsonNAmes

Cerchie:
## Summary of Changes

There was a test failing on main. 
Screenshot 2026-06-30 at 1 32 18 PM

Turns out we had a sneaky casing issue in the schema registry helper itself, see the comment:

```
/**
 * Sets each field's `jsonName` to the standard lowerCamelCase form of its
 * (possibly snake_case) name, recursing into nested message types.
 *
 * protobufjs's `Field#toDescriptor` never populates `jsonName` (see its
 * "Not supported" doc comment), so a descriptor built from `.proto` text always
 * has it unset. `@bufbuild/protobuf` takes `jsonName` as authoritative — it
 * does not fall back to deriving it from `name` — so without this, `fromJson`
 * only accepts the literal proto field name and rejects the camelCase JSON
 * name a real protoc-generated descriptor would also accept.
 */
```

N/A, can run tests locally to confirm 

## Pull request checklist

#### Tests

- [ ] Added new
- [x] Updated existing-- technically, since this casing change fixes the test. 
- [ ] Deleted existing

### DIFF
--- a/src/confluent/schema-registry-helper.test.ts
+++ b/src/confluent/schema-registry-helper.test.ts
@@ -296,6 +296,55 @@ describe("schema-registry-helper.ts", () => {
       });
     });
 
+    it("accepts camelCase payload keys on the use-latest path (proto json_name)", async () => {
+      const registry = newRegistry();
+      const topic = "proto-latest-camelcase";
+      // Seed the subject the same way: the first produce supplies the schema,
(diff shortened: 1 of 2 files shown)


Cerchie:
closing, no longer failing on main

Issue #367, opened 2026-05-06
Thread `connectionId` through tool handlers that depend on connection-specific config


shouples:
### Problem

The runtime currently assumes a single connection. `runtime.clientManager` (https://github.com/confluentinc/mcp-confluent/blob/1a3cd3c/src/server-runtime.ts) throws on 0 or >1 managers, and many handlers reach for `runtime.requireDirectClientManager()` or `getSoleDirectConnection()`. With multi-connection support arriving as the next milestone after #151, every tool that operates on connection-specific resources (Kafka topics, Flink statements, Schema Registry, etc.) needs a way for the caller to specify *which* connection to target.

### Proposed design (initial sketch - expected to be broken down further)

1. Add `runtime.getClientManager(connectionId: string): BaseClientManager` (or similar id-based lookup) to https://github.com/confluentinc/mcp-confluent/blob/1a3cd3c/src/server-runtime.ts.
2. Audit handlers under https://github.com/confluentinc/mcp-confluent/blob/1a3cd3c/src/confluent/tools/handlers/. For each handler that depends on a specific connection's config:
   - Add `connectionId: string` to the input schema.
   - Replace `runtime.clientManager` / `requireDirectClientManager()` with the new id-based lookup.
3. Decide the routing UX:
   - **Required `connectionId`** is safest - no break-on-promotion behavior when a second connection is configured (matches the decision made for #365).
   - **Optional with sole-connection fallback** is friendlier in single-connection setups but breaks once a second connection lands.

### Open questions

- Per-handler decision on required vs optional `connectionId`, or a project-wide policy?
- Migration cadence: one mega-PR auditing every handler, or a domain-by-domain rollout (kafka, then flink, then schema, etc.)?
- Failure modes when a handler is called with an id whose connection lacks the required block (e.g., `list-flink-statements` against a connection with no `flink:` block) - error in the handler, or rely on `enabledConnectionIds` + #358 so clients never call it that way?

This issue is intentionally rough and will be broken down into smaller children as the design firms up.

### Relationships

- Tracks against: #151 (single-connection YAML migration) and the planned multi-connection epic that follows it.
- Becomes load-bearing for: #364, #365, #366 (the discovery / config-help track) - those tools can ship before this one, but their value lights up once routing lands.

ramboweb3-rgb:
[...]

PR #682, opened 2026-07-07
Upgrade kafka-javascript; remove wrong typing workarounds

### LINKED ISSUE #514: Remove workarounds and tests over workarounds when the next confluent-kafka-javascript release happens
Dependency confluent-kafka-javascript had various typescript typing issues that we had to work around. When its next release happens and we upgrade to it, then look to undo the workarounds we did for the issues:

  * https://github.com/confluentinc/confluent-kafka-javascript/issues/367
  * https://github.com/confluentinc/confluent-kafka-javascript/issues/487
  * https://github.com/confluentinc/confluent-kafka-javascript/issues/489

jlrobins:
## Upgrade @confluentinc/kafka-javascript, drop typing issues workarounds

- Bumped `@confluentinc/kafka-javascript` specification from `1.3.2` to `1.10.0` (in reality, going from `1.9.0` to `1.10.0`). The three upstream type/runtime mismatches we'd worked around (confluentinc/confluent-kafka-javascript#367, #487, #489) were all fixed in `1.9.1`; `1.10.0` is the current latest and carries no breaking changes over `1.3.2` per the library's changelog.
- Deleted `topic-metadata.ts`'s `normalizeFetchTopicMetadataResponse()` (and its test file) now that `admin.fetchTopicMetadata()` is correctly typed as `Promise<Array<ITopicMetadata>>` (#367 fix) — call sites in `get-consumer-group-lag-handler.ts` and `consume-kafka-messages-handler.ts` consume the array directly.
- Simplified `describe-consumer-group-handler.ts`'s `buildAssignment()` to a single `member.assignment.topicPartitions.map(...)` — the dual-shape defensive branch (flat array vs. `{topicPartitions}` wrapper) and its "Wacky" fallback log are gone now that the upstream `MemberDescription.assignment` type honestly declares the wrapper (#487 fix; upstream even adopted the `MemberAssignment` named type and `Buffer | null` suggestion from our issue write-up verbatim).
- Dropped the loose-equality (`!= null`) workaround for `FetchOffsetsPartition.error` in `get-consumer-group-lag-handler.ts` in favor of a strict `!== null` check — the type is now honestly `LibrdKafkaError | null` (#489 fix), so `undefined` is no longer part of its domain.
- Left `consumer-group-helpers.ts`'s `GroupDescription.error != null` loose-equality check untouched: that field is still typed `LibrdKafkaError | undefined` upstream while runtime populates `null`, a distinct, still-open mismatch not covered by #514's tracked issues (367/487/489).
- Verified via `tsc --noEmit`, `pnpm run lint`, and `pnpm run test:unit` (128 files / 2408 tests, all green) plus a diff-line coverage check — every changed line in the four touched handler/helper files is covered.

## Relationships

Closes #514


flippingbits (approved):
Thanks! I believe a change log entry shouldn't hurt but I don't want to block the PR on it.

PR #354, opened 2026-05-05
Add get_product_doc_page tool

eguicf:
## Summary of Changes

- This PR adds a new `get-product-doc-page` MCP tool that fetches the full markdown content of a single Confluent product documentation page from `docs.confluent.io`, `developer.confluent.io`, or `support.confluent.io`.
- Companion to `search-product-docs` (PR #267): search returns URLs, this tool reads them.
- Closes https://confluentinc.atlassian.net/browse/DTX-978 ([research note](https://confluentinc.atlassian.net/wiki/spaces/DTX/pages/5508923513?themeState=dark%3Adark+light%3Alight+spacing%3Aspacing+typography%3Atypography+colorMode%3Alight))

#### Per-host fetch strategy: each site exposes content differently

Host | Primary | Fallback | Why
-- | -- | -- | --
docs.confluent.io | .md twin URL (200 + content-type: */markdown) | div.rst-content → <body> | Sphinx publishes .md source alongside .html
developer.confluent.io | [data-swiftype-name="body"] | section[data-test-id="section"] → stripped <body> | Gatsby hashes CSS classes; only data-* attrs are stable
support.confluent.io | Zendesk JSON API: /api/v2/help_center/articles/<id>.json | article#main-content | Public API returns body without theme chrome

All HTML is converted to markdown via Turndown.

### Manual testing instructions

**Build**
```
npm run build
```

**Run the tool**

```
# Track A: Inspector UI
npm run inspector
```
```
#Track B: Claude Code (and other AI assistants)
claude mcp add mcp-confluent-dev -- node "$(pwd)/dist/index.js" --env-file "$(pwd)/.env"
claude mcp list
```
**In the Inspector UI:**

1. Tool listing
   confirm `get-product-doc-page` appears with required `url: string`
2. Test `docs.confluent.io`
   input: https://docs.confluent.io/cloud/current/get-started/
   expect: markdown body starting with "# Quick Start for Confluent Cloud"
3. `developer.confluent.io` — Swiftype body
   input: https://developer.confluent.io/courses/apache-iceberg/introduction/
   expect: lesson body, no sidebar/nav
4. `developer.confluent.io` — page-data full page
   https://developer.confluent.io/get-started/java/
   → contains "In this tutorial", "gradle shadowJar"
5. `developer.confluent.io` — page-data fragment
   https://developer.confluent.io/get-started/java/#produce-events
   → ONLY the Produce Events section
   → does NOT contain "Prerequisites" or "In this tutorial"
[...]

PR #706, opened 2026-07-17
[Prototype for evaluation] Replace `turndown` with a small `cheerio`-based converter in `get-product-doc-page`

eguicf:
## Summary of Changes

- `turndown` depends on `@mixmark-io/domino` as its Node.js HTML parser. Some
  enterprise Artifactory/security scanners quarantine `@mixmark-io/domino`,
  which breaks `npm install -g @confluentinc/mcp-confluent` for customers on
  locked-down networks. ([thread](https://confluent.slack.com/archives/C0AJUENFYDC/p1784236880201669))
- This PR writes a small converter (`html-to-markdown.ts`, ~185 lines) using
  `cheerio`, which is already a direct dependency of this project. No new
  package is introduced at all, so this removes the whole class of
  "third-party HTML parser gets flagged" risk rather than trading one
  flagged dependency for a different, less-scrutinized one.
- Covers the specific tag set Confluent doc pages actually use (headings,
  paragraphs, bold/italic, links, images, inline/fenced code, ordered/
  unordered lists incl. nesting, blockquotes, tables), based on real-page
  samples pulled via `search-product-docs` during evaluation. Unknown tags
  fall back to plain text.
- Bonus: renders real Markdown tables. `turndown` (without the GFM plugin,
  which this project doesn't use) has no table support at all, it dumps
  each cell's text on its own line with no row/column structure.
- Known limitations (not hit on any real page tested): nested tables aren't
  handled correctly, and a backtick inside inline `<code>` text isn't escaped.
- No call-site changes outside `get-product-doc-page-handler.ts`; `turndown`
  was only used there.

## Test plan
- [x] `pnpm run typecheck` — passes
- [x] `pnpm run lint` — passes
- [x] `pnpm run build` — passes
- [x] `pnpm run test:unit` — 2429/2429 passes
- [x] `grep -c "turndown\|node-html-markdown\|@mixmark-io/domino" pnpm-lock.yaml` → `0`
- [x] Manually verified against real pages on all three hosts
      (`docs.confluent.io`, `developer.confluent.io`, `support.confluent.io`)

### Manual Testing Instructions

refer to https://github.com/confluentinc/mcp-confluent/pull/354

Additionally worth checking against this PR specifically:
1. `https://developer.confluent.io/what-is-apache-kafka/`
   → link text/hrefs are preserved even where the source HTML has an
     `<a>` sitting directly inside a `<div>` with no wrapping `<p>`
   → no stray `[]()` artifacts from heading permalink icons
2. Any support.confluent.io article containing a table (e.g. a release-notes
   article) → renders as a proper `| a | b |` Markdown table, not flattened
   cell-by-cell text

#### Tests

- [ ] Added new
- [ ] Updated existing
- [ ] Deleted existing

#### Release notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,6 @@
     "@types/content-type": "^1.1.9",
     "@types/eslint__js": "^8.42.3",
     "@types/node": "^24.0.4",
-    "@types/turndown": "^5.0.6",
     "@types/ws": "^8.18.1",
     "@typescript-eslint/eslint-plugin": "^8.35.0",
     "@typescript-eslint/parser": "^8.35.0",
@@ -96,7 +95,6 @@
     "pino-pretty": "^13.1.3",
     "properties-file": "^3.5.12",
     "protobufjs": "^8.6.3",
-    "turndown": "^7.2.4",
     "yaml": "^2.8.3",
     "zod": "^4.0"
   },
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@typescript-eslint/parser": "^8.35.0",
     "@vitest/coverage-v8": "^4.1.0",
     "concurrently": "^9.2.0",
+    "domhandler": "5.0.3",
     "eslint": "^9.29.0",
     "eslint-config-prettier": "^10.1.5",
     "eslint-plugin-prettier": "^5.5.1",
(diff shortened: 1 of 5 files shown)


eguicf:
Closing — customer issue resolved.

PR #724, opened 2026-08-20
(main) Fix PROTOBUF produce fails on .proto-text-registered subjects (#720)

### LINKED ISSUE #720: PROTOBUF produce fails on .proto-text-registered subjects (use-latest path)
Producing to a PROTOBUF subject fails when the subject was registered as `.proto` text via the standard Schema Registry REST API. It only works when the schema was registered by the Confluent JS SDK's own serializer (base64 `FileDescriptorProto`). AVRO and consuming work.

**Root cause:** In `src/confluent/schema-registry-helper.ts`, the use-latest branch of `serializeProtobufMessage` always calls `protobufRegistryFromSerialized`, which base64-decodes the stored schema as a `FileDescriptorProto`. For a `.proto`-text subject the stored value is literal proto source, so `Buffer.from(text, "base64")` yields garbage and `fromBinary(...)` throws (`Failed to decode registered Protobuf schema`). AVRO is unaffected — it lets the SDK consume the schema string directly. The correct `.proto` parser (`protobufRegistryFromProto`) already exists but is only used on the caller-supplied-schema path.

**Fix:** In the use-latest branch, dispatch on the stored schema format (or try `protobufRegistryFromSerialized`, fall back to `protobufRegistryFromProto`).

**Tests:** Add a use-latest PROTOBUF case seeded with a `.proto`-text subject

Reported from https://softwaremill.com/managing-kafka-with-ai-and-mcp/

jlrobins:
## Summary of Changes

Fix for #720 against main:

- **`produce-message` (PROTOBUF, use-latest).** Producing to a PROTOBUF subject registered as `.proto` text via the standard Schema Registry REST API no longer fails with "Failed to decode registered Protobuf schema".

New integration test passes:

 ✓  integration  src/confluent/tools/handlers/kafka/produce-kafka-message-handler.integration.test.ts > produce-kafka-message-handler > with a Protobuf subject registered as .proto text via the REST API (issue #720) > via stdio transport > should produce via the use-latest path and decode back through consume-messages 17074ms

Closes #720

### DIFF
--- a/src/confluent/tools/handlers/kafka/produce-kafka-message-handler.integration.test.ts
+++ b/src/confluent/tools/handlers/kafka/produce-kafka-message-handler.integration.test.ts
@@ -551,5 +551,112 @@ describe(
         });
       },
     );
+
+    describe(
+      "with a Protobuf subject registered as .proto text via the REST API (issue #720)",
+      { tags: [Tag.REQUIRES_SCHEMA_REGISTRY_CONFIG] },
+      () => {
+        if (!activeConnectionTypes.includes(ConnectionType.DIRECT)) {
+          it.skip(CONNECTION_TYPE_DIRECT_FILTERED_REASON, () => {});
+          return;
+        }
+        if (skipIfDisabled(handler, integrationConnection())) {
+          return;
+        }
+        const connection = integrationConnection();
+        if (connection.type !== "direct" || !connection.schema_registry) {
+          it.skip("requires schema_registry config", () => {});
+          return;
+        }
+
+        // The standard Schema Registry REST API stores a Protobuf schema as
+        // literal .proto source text, distinct from the base64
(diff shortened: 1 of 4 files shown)


src/confluent/tools/handlers/kafka/produce-kafka-message-handler.integration.test.ts
```diff
@@ -551,5 +552,118 @@ describe(
...
+          return;
+        }
+        const connection = integrationConnection();
+        if (connection.type !== "direct" || !connection.schema_registry) {
+          it.skip("requires schema_registry config", () => {});
+          return;
+        }
+
+        // The standard Schema Registry REST API stores a Protobuf schema as
+        // literal .proto source text, distinct from the base64
```
Cerchie, line 574: Did this manually, LGTM

Cerchie (approved):

PR #734, opened 2026-09-21
Stop reporting caller-fixable tool errors to Sentry

bcscc:
## Summary

Discovered while click-testing 1.6.0: `createMcpServer`'s tool-call wrapper (`src/mcp/server.ts`) reports **every** thrown tool-call error to Sentry indiscriminately, including expected, caller-correctable input mistakes — a bad param (`ZodError` from `schema.parse()`) or a required id/config value missing under OAuth or a bare direct connection. Since a bad-params call from an LLM caller guessing at arguments is expected, high-volume traffic (not a server fault), this would be very noisy in production if left as-is.

- Introduces `ToolInputError` (`src/confluent/tools/tool-input-error.ts`) — thrown instead of a plain `Error` for caller-correctable input errors.
- Converted the throw sites that fit this shape:
  - `BaseToolHandler.resolveParam` and all of `cluster-arg-resolvers.ts`'s resolvers (missing required arg/id, with or without a config fallback).
  - `FlinkToolHandler.resolveFlinkRouting`'s OAuth-required-ids check.
  - `consume-kafka-messages-handler.ts`: 9 of its 11 validation throws (per-topic `start`/`partition` conflicts, out-of-range offset/tail/timestamp, nonexistent topic). The remaining 2 (watermark-lookup returning nothing for an already-validated partition) are left as plain `Error` — that's an internal consistency issue, not a caller mistake.
  - `get-connector-logs-handler.ts`'s missing `confluent_cloud.auth` check.
- `createMcpServer`'s tool-call wrapper now skips `captureError` when the thrown error is a `ZodError` or `ToolInputError`, via a new `isExpectedToolError()` helper. The error is still `throw`n and still surfaces to the caller as a normal `isError: true` tool result with the same message as before — this change only affects Sentry reporting, not caller-facing behavior.

Left unconverted (genuine faults, correctly still reported): upstream-API-response validation in `list-clusters-handler.ts`/`list-compute-pools-handler.ts`, telemetry-API-error throws in the metrics handlers, exhaustiveness guards in `describe-fields.ts`/`config-help-handler.ts`, and the docs-fetch timeout/HTTP-status throws in `docs/*` (mixed — a non-existent doc URL and an upstream 5xx currently share the same throw; flagged but not changed here).

## Test plan
- [x] `pnpm run typecheck` clean
- [x] `pnpm run lint` clean
- [x] New `src/confluent/tools/tool-input-error.test.ts` covers `ToolInputError` and `isExpectedToolError()`
- [x] Existing unit suites for every touched file pass (`base-tools`, `cluster-arg-resolvers`, `flink-tool-handler`, `consume-kafka-messages-handler`, `get-connector-logs-handler`, `mcp/server`) — 240 tests green


eguicf (approved):
Thanks for the PR, LGTM

